PR #116, opened 2024-03-12
Fix internal link in chart docs

andresgutgon:
# What?
Small fix ocd


packages/client/svelte/src/lib/ui/select/select-content.svelte
andresgutgon, line 1: Just the prettier config when I saved this file I think

PR #130, opened 2024-03-14
feature: Alert component

geclos:
# WHAT

Implements the shadcn alert component in our svelte component lib


andresgutgon (dismissed):

packages/client/core/src/theme/ui/alert/index.ts
```diff
@@ -0,0 +1,35 @@
...
+function rootCssClass({
+  variant,
+  className,
+}: {
+  variant: Variant
+  className?: string | null
+}) {
+  return cn(
+    alertVariants({ variant }),
+    'text-sm [&_p]:leading-relaxed',
```
geclos, line 30: all alerts have this text styles which will conflict if users try to add other components inside the alert but we prefer a super simple api over extensibility in this case

csansoon (approved):

PR #131, opened 2024-03-14
minor: added ssl support to postgres connector

geclos:
# WHAT
Implements ssl support for the postgres connector

### DIFF
--- a/packages/connectors/postgresql/src/index.ts
+++ b/packages/connectors/postgresql/src/index.ts
@@ -6,7 +6,16 @@ import {
   ResolvedParam,
 } from '@latitude-data/base-connector'
 import QueryResult, { DataType } from '@latitude-data/query_result'
-import pg from 'pg'
+import { DatabaseError, Pool, types as pgtypes } from 'pg'
+import { readFileSync } from 'fs'
+
+type SSLConfig = {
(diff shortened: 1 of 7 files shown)


packages/connectors/postgresql/src/index.ts
```diff
@@ -84,36 +94,68 @@ export class PostgresConnector extends BaseConnector {
...
+      return ssl
+    }
+
+    const { sslmode, ca, key, cert, rejectUnauthorized } = ssl
+    const payload = {
+      sslmode,
+      rejectUnauthorized,
+      ca: ca ? readFileSync(ca).toString() : undefined,
+      key: key ? readFileSync(key).toString() : undefined,
+      cert: cert ? readFileSync(cert).toString() : undefined,
```
andresgutgon, line 153: Where are these files defined?

geclos: the client passes the paths to the files themselves

csansoon: What if those files do not exist? Is there an understandable error message explaining why it failed?

geclos: good point we should handle this gracefully

csansoon (approved):

PR #133, opened 2024-03-15
feature: add automatic releases from CI

geclos:
# WHAT
Adds necessary configuration for automatic releases using changesets github action. This action will also automatically create the github release tags upon publishing. Here are their [docs](https://github.com/changesets/action?tab=readme-ov-file#inputs).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,7 +30,12 @@ jobs:
           version: 8
           run_install: true
 
-      - name: Create Release Pull Request
+      - name: Create Release Pull Request or Publish to npm
+        id: changesets
         uses: changesets/action@v1
+        with:
+          # This expects us to have a script called release which does a build for our packages and calls changeset publish
+          publish: pnpm release
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -11,6 +11,7 @@
     "prettier": "turbo prettier",
     "build": "turbo build",
     "format": "prettier --write \"**/*.{ts,tsx,md}\"",
+    "release": "pnpm run build && pnpm run changeset publish",
```
geclos, line 14: I think the run build prestep is not really needed because we have prepublish steps on all package.json files, but just to be safe I'm adding it for now. I'll remove it if I see it's not really needed.

.github/workflows/release.yml
```diff
@@ -30,7 +30,17 @@ jobs:
...
+          # This expects you to have a script called release which does a build for your packages and calls changeset publish
+          publish: pnpm release
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      - name: Send a Slack notification if a publish happens
+        if: steps.changesets.outputs.published == 'true'
+        # You can do something when a publish happens.
+        run: my-slack-bot send-notification --message "A new version of ${GITHUB_REPOSITORY} was published!"
```
csansoon, line 46: Does GitHub Actions recognise `my-slack-bot` automatically, or do we have to configure it somewhere?

geclos: we have to configure it estoy en ello

PR #136, opened 2024-03-15
This is a CI test

geclos:
DO NOT MERGE

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -60,4 +60,4 @@ jobs:
         run: pnpm turbo run build --cache-dir=.turbo
 
       - name: Run tests
-        run: pnpm turbo run test --filter=[HEAD^1] --filter=[HEAD^1]
+        run: pnpm turbo run test --filter=[HEAD^1]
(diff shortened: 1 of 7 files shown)


.github/workflows/test.yml
```diff
@@ -58,7 +58,7 @@ jobs:
 
           # tells Turbo to only run the script (build or test in this case) for the packages
       - name: Use Turbo to build affected packages
-        run: pnpm turbo run build --cache-dir=.turbo --filter=@latitude-data/*{./packages/*}[HEAD^1] --filter=@latitude-data/*{./apps/*}[HEAD^1]
+        run: pnpm turbo run build --cache-dir=.turbo --filter=./packages/* --filter=./apps/*
 
       - name: Run tests
-        run: pnpm turbo run test --filter=@latitude-data/*{./packages/*}[HEAD^1] --filter=@latitude-data/*{./apps/*}[HEAD^1]
+        run: pnpm turbo run test --filter=./packages/* --filter=./apps/*
```
andresgutgon, line 64: Without the git part this is not useful. It will run all the packages. Not only the ones that changed since last commit I think

geclos: tranquilo que estoy haciendo pruebas

PR #139, opened 2024-03-15
minor: ignore cli spawn process warnings

geclos:
They are harmless and only increase noise


geclos:
🐮 👦🏼 since it's a super small change

PR #147, opened 2024-03-19
Alert component redesign

andresgutgon:
# What?
Change styles on our `<Alert />` component to much more Latitude design system.

### TODO
- [x] Copy styles from old `<Notification />` component
- [x] Make sure a chart error looks good

image


packages/client/svelte/src/lib/ui/alert/alert.svelte
```diff
@@ -1,23 +1,44 @@
...
+  export let secondary: $$Props['secondary'] = false
   export { className as class }
+
+  $: styles = theme.ui.alert.cssClass({
+    type,
+    scrollable,
+    secondary,
+    className,
+  })
+  $: textColor = styles.properties.foreground
```
andresgutgon, line 33: If the prop `type` change `cssClass` compute new `background` and `foreground` colors. But `textColors` is not updated. Any idea why?

packages/client/svelte/src/lib/ui/alert/alert.svelte
```diff
@@ -1,23 +1,45 @@
...
   import { theme } from '@latitude-data/client'
-  import type { HTMLAttributes } from 'svelte/elements'
+  import AlertText from './text.svelte'
 
-  type $$Props = HTMLAttributes<HTMLDivElement> & theme.ui.alert.Props
+  type $$Props = Props
 
   let className: $$Props['class'] = undefined
-  export let variant: $$Props['variant'] = 'default'
+  export let type: $$Props['type']
```
csansoon, line 20: Update the Docs with the changes on API

csansoon (dismissed):

packages/client/svelte/src/lib/ui/alert/alert.svelte
andresgutgon: Good catch 💯

packages/client/svelte/src/lib/ui/alert/alert.svelte
geclos: :\

PR #148, opened 2024-03-19
feature: sync dependencies with Latitude

geclos:
# WHAT

Sync dependencies between project's root package.json and latitude's apps/server package.json. This allows users to declare and install dependencies from the root of the project. The main tradeoff is that potential dependency conflicts might happen, so we have to warn the user about those manually.


packages/cli/src/lib/rootPath.ts
```diff
@@ -1,11 +0,0 @@
...
-import config from '../config'
-import { DEV_SITES_ROUTE_PREFIX } from '../commands/constants'
-
-// Returns the root path of your data app. This is used to prefix the routes in
-// development whereas in production it will always be /.
-export default function rootPath() {
-  return config.pro
-    ? ''
-    : `/${DEV_SITES_ROUTE_PREFIX}/${path.basename(config.cwd)}`
-}
```
andresgutgon, line 11: Why you're removing this?

geclos: it conflicts with package.json in the `sites/dev-*` folder now, I'm trying to figure out another solution, that also does not involve a custom folder structure hardcoded in code

andresgutgon: What's `sites/dev-*`?

geclos: the projects that you install with latitude-dev start

geclos: I want to make it so the dev environment is not in a subroute

andresgutgon: But then it would be committed to the repo no?

geclos: let's talk offline

packages/cli/src/lib/setupApp/index.ts
```diff
@@ -1,54 +1,72 @@
...
-  let allGood = false
 
+// Adds a package.json file to the app directory in development environment so
+// that you can run the cli in development mode with `npm run latitude-dev`
+function addPackageJson() {
+  const packageJson = {
+    scripts: {
+      'latitude-dev': '../../node_modules/.bin/latitude',
+    },
+  }
```
csansoon, line 19: Where is this supposed to be ran from?

csansoon: Oh, I see. Nevermind

geclos: from the project directory. This is only for development. So from your project directory you can do `npm run latitude-dev <whatever command you want>`

csansoon: Since the `latitude` script is in the outer `node_modules`, why not just run `npm run latitude-dev` from the root directory, instead of the inner app_folder?

csansoon (dismissed):

packages/cli/src/commands/start/index.ts
```diff
@@ -29,7 +29,7 @@ async function displayMessage() {
     --------------------------------------
 
     $ cd ./${projectDir.includes(' ') ? `"${projectDir}"` : projectDir}
-    $ ${config.dev ? 'pnpm latitude-dev dev' : 'latitude dev'}
+    $ ${config.dev ? 'npm run latitude-dev' : 'latitude dev'}
```
andresgutgon, line 32: Use configured package manager. Check config

andresgutgon: Wait, this is for development. Why `npm` and not pnpm`?

geclos: well could be pnpm or npm or whatever npm package you wanna use 😅

andresgutgon: But we (developers of monorepo) all use `pnpm`

geclos: ok pnpm then

andresgutgon (dismissed):

.changeset/itchy-bananas-yawn.md
```diff
@@ -0,0 +1,7 @@
+---
+"@latitude-data/cli": minor
+"@latitude-data/connector-factory": patch
+---
+
+CLI (minor): Allow users to install custom npm dependencies to their Latitude projects with regular npm install
+Connectors factory (patch): Raise an error if user tries to use a secret environment variable that we cannot find in the environment.
```
andresgutgon, line 7: Add documentation

geclos: aaah good catch!

PR #154, opened 2024-03-20
Fix: queries not refetching

csansoon:
Solved problems:
 - RunButton didn't call any refetch if no queries were specified
 - All initial fetches were using `force: true`
 - The force tag wasn't being added to force requests

Also changed the method of sending the force tag from a `__force` parameter to a `X-Force-Refetch` header.

### DIFF
--- a/apps/server/src/lib/stores/queries.ts
+++ b/apps/server/src/lib/stores/queries.ts
@@ -69,10 +69,12 @@ async function fetchQueryFromCore({
   query,
   inlineParams,
   force = false,
+  skipIfParamsUnchanged = true,
 }: {
   query: string
   inlineParams: InlineParams
-  force?: boolean
+  force?: boolean // Adds the 'force' flag to the request, to invalidate the backend cache
+  skipIfParamsUnchanged?: boolean // If true, it won't refetch if the params haven't changed
 }): Promise<void> {
   const queryKey = createMiddlewareKey(query, inlineParams)
   const computedParams = computeQueryParams(inlineParams)
@@ -86,7 +88,7 @@ async function fetchQueryFromCore({
   }))
 
   if (!browser || !loaded) return // Don't fetch queries until the page is fully loaded
-  if (!force && oldQueryKey === coreQueryKey) return // Don't refetch if there have been no changes to the params (unless force is true)
+  if (skipIfParamsUnchanged && oldQueryKey === coreQueryKey) return // Don't refetch if there have been no changes to the params
 
   if (force) {
     queryStore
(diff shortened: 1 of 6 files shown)


apps/server/src/lib/stores/queries.ts
```diff
@@ -69,10 +69,12 @@ async function fetchQueryFromCore({
   query,
   inlineParams,
   force = false,
+  ignoreLoadTag = false,
 }: {
   query: string
   inlineParams: InlineParams
   force?: boolean
+  ignoreLoadTag?: boolean
```
andresgutgon, line 77: Can you explain this param? Maybe a comment on top of this type

packages/client/core/src/stores/queries.ts
```diff
@@ -97,7 +99,8 @@ export const store = createStore<StoreState>((set, get) => {
         // TODO: Add force parameter or header when backend cache is implemented
         const response = await api.get<QueryResultPayload>(
           `api/queries/${queryPath}`,
-          params
+          params,
+          { [FORCE_REFETCH_HEADER_NAME]: 'true' }
```
geclos, line 103: always prefix any custom header with X-

packages/client/core/src/stores/queries.ts
```diff
@@ -4,6 +4,8 @@ import QueryResult, {
   type QueryResultPayload,
 } from '@latitude-data/query_result'
 
+export const FORCE_REFETCH_HEADER_NAME = 'Force-Refetch'
```
geclos, line 7: Prefix the custom http header with `X-`

csansoon: The `X-` convention is outdated and [is no longer needed](https://stackoverflow.com/questions/3561381/custom-http-headers-naming-conventions) for custom headers.
I'll add it anyways though.

apps/server/src/lib/stores/queries.ts
csansoon: I renamed it to be more clear. What do you think about it now?

geclos: better!

apps/server/src/lib/stores/queries.ts
```diff
@@ -69,10 +69,12 @@ async function fetchQueryFromCore({
   query,
   inlineParams,
   force = false,
+  skipIfParamsUnchanged = true,
 }: {
   query: string
   inlineParams: InlineParams
-  force?: boolean
+  force?: boolean // Adds the 'force' flag to the request, to invalidate the backend cache
+  skipIfParamsUnchanged?: boolean // If true, it won't refetch if the params haven't changed
```
andresgutgon, line 77: Nice, clear

geclos: why is it we need this flag? in what instance do we wanna fetch if the params have not changed?

csansoon: 1. When forcing
2. When initializing, due to the "not fetch until all params have loaded", any fetch in this phase is actually being skipped. Then, when the view has loaded, all queries are ran, but since the params have not changed then the fetch request is not sent. That's why I added this flag.

PR #160, opened 2024-03-21
fix: incorrect result parsing in snowflake connector

geclos:
# WHAT

Fixes a bug that caused snowflake results to not get correctly parsed in most queries.


geclos:
🐮 👦🏼 very small change

PR #163, opened 2024-03-21
Refactor CLI config to allow for non-singleton config

csansoon:
The VSCode extension should use functions and libraries defined in the CLI package to configure, run, manage files and projects.

The current implementation of CLI is using a `config` singleton that that is initiated from the terminal cwd. The expected use for VSCode, instead, could have different files from different Latitude projects opened in the same Editor, and it should handle them with their own config accordingly, which does not have to be from the workspace source either.

To fix this, now some CLI functions may accept a custom `CLIConfig` class, which will use the singleton by default.


packages/cli/rollup.config.mjs
```diff
@@ -23,11 +23,15 @@ const packageVersion =
...
 export default {
-  input: 'src/index.ts',
+  input: {
+    index: 'src/index.ts',
+    cli: 'src/cli.ts',
+  },
   cache: false, // This is generating stale builds in watch mode
   output: {
-    file: 'dist/index.js',
+    dir: 'dist',
```
andresgutgon, line 32: Check `dist` folder looks good, that types are exported correctly

packages/cli/src/cli.ts
```diff
@@ -0,0 +1,90 @@
...
+  .option('--watch', 'Re-run the query each time the query file changes')
+  .option(
+    '--param',
+    'Add a parameter to the query. Use the format --param <name>=<value>',
+  )
+  .example('run --watch users')
+  .example('run users --param user_id=foo')
+  .action(command('runCommand'))
+
+function cliFactory() {
```
andresgutgon, line 55: Put this on top. I always "flipo" went I start reading a file and don't see the import or the function

csansoon: Yeah I was wondering why is this even defined in a function

packages/cli/src/lib/maybeSetupApp/index.ts
```diff
@@ -1,11 +1,11 @@
-import config from '$src/config'
+import { CLIConfig } from '$src/config'
 import fsExtra from 'fs-extra'
 import path from 'path'
 import setupApp from '../setupApp'
 import { LATITUDE_FOLDER } from '$src/commands/constants'
 
-export default async function maybeSetupApp() {
-  const hasApp = fsExtra.existsSync(path.join(config.cwd, LATITUDE_FOLDER))
+export default async function maybeSetupApp(config: CLIConfig = CLIConfig.getInstance()) {
```
andresgutgon, line 7: Ok so what this is doing is getting the singleton as default and you're setting `config` to pass from the VSCode extension?

packages/cli/src/cli.ts
andresgutgon: No need right?

packages/cli/rollup.config.mjs
csansoon: Hmm, good catch! Now `src/index.ts` and `src/cli.ts` are indeed being compiled into `dist/index.js` and `dist/cli.js` respectively, but the type files are being exported to `dist/src/*.d.ts`.

Would changing the package.json to `types: "dist/src/index.d.ts` work? Or do I have to change the tsconfig.json "rootDir"?

packages/cli/src/lib/maybeSetupApp/index.ts
csansoon: The VSCode extension will create a custom `config` depending on the open Latitude file, with its own `source` directory. Currently I am not using `maybeSetupApp`, but it just defaults to the config singleton.

packages/cli/src/cli.ts
```diff
@@ -0,0 +1,85 @@
...
+  .option(
+    '--param',
+    'Add a parameter to the query. Use the format --param <name>=<value>',
+  )
+  .example('run --watch users')
+  .example('run users --param user_id=foo')
+  .action(command('runCommand'))
+
+
+async function init() {
```
andresgutgon, line 67: 👾

andresgutgon (dismissed):
Do QA, run all commands with all the combinations

andresgutgon (approved):

Issue #153, opened 2024-03-20
Notify new CLI versions


samulatitude:
Currently the Latitude CLI update is a manual `npm update -g @latitude-data/cli`, it doesn't notify when there is a new version and may cause bugs or incompatibilities. We should notify users of new CLI versions when they use a command.

andresgutgon:
This is what Storybook does. Would be something like this?
<img width="897" alt="image" src="https://github.com/latitude-dev/latitude/assets/49499/ece07c9b-b7aa-4bfb-937d-d3e9fa32cdd6">

geclos:
Makes sense! We should also handle different messages for each type of version, if it's minor or patch it's ok to strongly suggest to do a latitude update. If it's a major version we should suggest they visit firs the changelog.

PR #166, opened 2024-03-21
feature: cli version check on any command

geclos:
# WHAT

Checks for cli version mismatches and displays a prompt to update.

image

Solves https://github.com/latitude-dev/latitude/issues/153

### DIFF
--- a/packages/cli/src/commands/constants.ts
+++ b/packages/cli/src/commands/constants.ts
@@ -1,8 +1,9 @@
-export const LATITUDE_GITHUB_SLUG = 'latitude-dev'
-export const LATITUDE_FOLDER = '.latitude'
-export const LATITUDE_CONFIG_FILE = 'latitude.json'
-export const APP_FOLDER = `${LATITUDE_FOLDER}/app`
 export const APP_SERVER_FOLDER = 'apps/server'
-export const PACKAGE_NAME = '@latitude-data/server'
+export const CLI_PACKAGE_NAME = '@latitude-data/cli'
 export const DEFAULT_VERSION_LIST = ['latest'] // If we fail to get the list
-export const LATITUDE_VERSION = process.env.PACKAGE_VERSION
+export const LATITUDE_CONFIG_FILE = 'latitude.json'
+export const LATITUDE_FOLDER = '.latitude'
+export const LATITUDE_GITHUB_SLUG = 'latitude-dev'
+export const PACKAGE_NAME = '@latitude-data/server'
+
+export const APP_FOLDER = `${LATITUDE_FOLDER}/app`
--- a/packages/cli/src/commands/index.ts
+++ b/packages/cli/src/commands/index.ts
@@ -9,6 +9,7 @@ import trackedDecorator from '../lib/decorators/tracked'
 import { TelemetryEventType } from '../lib/telemetry/events'
 import { onError } from '../utils'
 import run from './run'
+import versionCheck from '$src/lib/decorators/versionCheck'
 
 export type Options = {
   tracked?: boolean
@@ -52,6 +53,7 @@ export default function command(
 
   if (tracked) commandFn = trackedDecorator(event, commandFn)
   if (setup) commandFn = setupDecorator(commandFn)
+  commandFn = versionCheck(commandFn)
 
   return commandFn
 }
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -1,6 +1,3 @@
-export * from './config';
+export * from './config'
 
-import sync from '$src/lib/sync'
-export {
-  sync,
-}
+export { default as sync } from '$src/lib/sync'
--- a/packages/cli/src/lib/decorators/versionCheck/index.ts
+++ b/packages/cli/src/lib/decorators/versionCheck/index.ts
@@ -1,0 +1,49 @@
+import boxedMessage from '$src/lib/boxedMessage'
+import chalk from 'chalk'
+import configStore from '$src/lib/configStore'
+import latestVersion from 'latest-version'
+import { CLIConfig } from '$src/config'
+import { CLI_PACKAGE_NAME } from '$src/commands/constants'
+
+const ONE_DAY = 1000 * 60 * 60 * 24
+
+export default function versionCheck(commandFn: Function) {
+  return async function (...args: any[]) {
+    try {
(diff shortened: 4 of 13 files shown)


packages/cli/src/lib/decorators/versionCheck/index.ts
```diff
@@ -0,0 +1,31 @@
...
+          title: 'Update available',
+          color: 'yellow',
+          textAlignment: 'center',
+          text: `
+A new version of @latitude-data/cli is available. 
+
+Current: ${chalk.red(packageJson.version)}
+Latest: ${chalk.green(latest)}
+
+Run ${chalk.cyan('npm i -g @latitude-data/cli')} to update.
```
andresgutgon, line 21: use `config.project.pkgManager.command` to avoid confusion

packages/cli/src/commands/constants.ts
```diff
@@ -1,8 +1,9 @@
-export const LATITUDE_GITHUB_SLUG = 'latitude-dev'
-export const LATITUDE_FOLDER = '.latitude'
-export const LATITUDE_CONFIG_FILE = 'latitude.json'
-export const APP_FOLDER = `${LATITUDE_FOLDER}/app`
 export const APP_SERVER_FOLDER = 'apps/server'
-export const PACKAGE_NAME = '@latitude-data/server'
+export const CLI_PACKAGE_NAME = '@latitude-data/cli'
 export const DEFAULT_VERSION_LIST = ['latest'] // If we fail to get the list
-export const LATITUDE_VERSION = process.env.PACKAGE_VERSION
```
geclos, line 8: removing this as it was not used afaict

andresgutgon: Here https://github.com/latitude-dev/latitude/blob/main/packages/cli/src/index.ts#L58

andresgutgon: Is generated at build time here https://github.com/latitude-dev/latitude/blob/main/packages/cli/rollup.config.mjs#L45

packages/cli/src/commands/index.ts
```diff
@@ -52,6 +53,7 @@ export default function command(
 
   if (tracked) commandFn = trackedDecorator(event, commandFn)
   if (setup) commandFn = setupDecorator(commandFn)
+  commandFn = versionCheck(commandFn)
```
andresgutgon, line 56: I concern I have with this is that you're doing a network request every time the users use the CLI 

We can go with this but I would try to do this:

Only check version when today is more than `lastTimeCliVersionCheckAt`. And you can store that in user's machine settings as we do with telemetry settings
https://github.com/latitude-dev/latitude/blob/main/packages/cli/src/lib/telemetry/index.ts#L68

``` typescript
this.configStore.set('lastTimeCliVersionCheckAt', new Date().toTimestamp());
```

Note: `toTimestamp` is invented but you get the idea

packages/cli/src/commands/constants.ts
geclos: :O ok

packages/cli/src/commands/index.ts
geclos: good point 👌🏼

packages/cli/src/commands/constants.ts
andresgutgon: What says your `git grep`?

packages/cli/src/index.ts
```diff
@@ -1,6 +1,3 @@
-export * from './config';
+export * from './config'
 
-import sync from '$src/lib/sync'
-export {
-  sync,
-}
+export { default as sync } from '$src/lib/sync'
```
andresgutgon, line 3: Before this was not default. Is this ok?

packages/cli/src/lib/decorators/versionCheck/index.ts
```diff
@@ -0,0 +1,49 @@
+import latestVersion from 'latest-version'
+import boxedMessage from '$src/lib/boxedMessage'
+import chalk from 'chalk'
+import { CLI_PACKAGE_NAME } from '$src/commands/constants'
+import { CLIConfig } from '$src/config'
+import Configstore from 'configstore'
+
+const config = new Configstore(CLI_PACKAGE_NAME, { lastCheckedAt: undefined })
```
andresgutgon, line 8: Instead of using `ConfigStore` outside couldn't you use from our config file? This way this package is not spread and it will be easier to replace if necessary and also to see all the keys that we're storing in user's config

packages/cli/src/lib/decorators/versionCheck/index.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export default function versionCheck(commandFn: Function) {
+  return async function (...args: any[]) {
+    try {
+      const lastCheckedAt = config.get('lastCheckedAt')
+      // return if checked less than a day ago
+      if (lastCheckedAt && Date.now() - lastCheckedAt < 24 * 60 * 60 * 1000) {
```
andresgutgon, line 15: Magic numbers. Do a constant please don't make calculate 😂

andresgutgon: i just saw the comment. More to my point. Better a constant than a comment

packages/cli/src/lib/telemetry/index.ts
```diff
@@ -142,15 +145,6 @@ class Telemetry {
...
   }
-
-  private get cliPkg() {
-    const packageJsonText = readFileSync(new URL('../package.json', import.meta.url))
-    const pkg = JSON.parse(packageJsonText.toString())
-    return {
-      name: pkg.name,
-      version: pkg.version
-    }
-  }
```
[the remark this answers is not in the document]
andresgutgon: I think so yes

packages/cli/src/index.ts
geclos: you are reading it wrong the rewrite I've done is functionally equivalent afaict

packages/cli/src/commands/constants.ts
geclos: i can't find anything that's my point

andresgutgon: So 🔥

andresgutgon (approved):

PR #167, opened 2024-03-22
Feature/detach query service from app

csansoon:
The functions in Query Service defined in the App are useful for another packages, like CLI and VSCode Extension, so I'm detaching it into its own package.


packages/query_service/.eslintrc.cjs
```diff
@@ -0,0 +1,7 @@
+/** @type {import("eslint").Linter.Config} */
+module.exports = {
+  extends: ['@latitude-data/eslint-config/library.js'],
+  parserOptions: {
+    project: true,
+  },
```
andresgutgon, line 6: no need for parseOptions I think

packages/query_service/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "prettier": "prettier --write src/**/*.ts"
+  },
+  "type": "module",
+  "main": "dist/index.ts",
+  "types": "dist/index.d.ts",
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "import": "./dist/index.js"
+    }
```
andresgutgon, line 21: In the case of a simple index no need for exports I think

packages/query_service/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "devDependencies": {
+    "@latitude-data/eslint-config": "workspace:*",
+    "@latitude-data/typescript": "workspace:*",
+    "@rollup/plugin-typescript": "^11.1.6",
+    "vitest": "^1.2.2",
+    "prettier": "^3.1.1",
+    "rollup": "^4.10.0"
+  },
+  "dependencies": {
+  }
```
andresgutgon, line 32: 🔥

packages/query_service/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": ["@latitude-data/typescript/base.json"],
+  "compilerOptions": {
+    "rootDir": "src",
+    "outDir": "dist"
```
andresgutgon, line 5: Missing root export types key

packages/query_service/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "@latitude-data/query_service",
+  "version": "0.0.1",
```
andresgutgon, line 3: I think you can start in 0.1.0

andresgutgon (dismissed):

PR #180, opened 2024-03-25
feature: split config responsibilities and simplify it

### LINKED ISSUE #185: Revert support to npm
From the get go we tried to support whatever package manager the user was using in Latitude's CLI: pnpm yarn or npm. The problem with this is that we either have to ask the user what manager they want the CLI to use or we have to take a guess, which more often than not could be wrong. In order to deliver a more robust solution and taking into consideration our bandwith we are gonna remove support for alternative package managers and instead focus on providing a rock solid solution with npm, which ships with all versions of NodeJS and is plenty fast for all versions of node that Latitude supports.

geclos:
# WHAT

Config is currently a complex state store that mutates throughout the code execution which makes it very hard to understand what is happening. I'm splitting its responsibilities into smaller subcomponents that will each perform a single, idempotent task whenever possible. This PR will be followed by more optimizations in the near future.

In this commit:
- Removed latitude json config from config global, now we get it from `await findOrCreateConfigFile`
- Made config file a module-level singleton
- Stop passing around config as prop everywhere, the point of it being a global is that you don't need to inject it everywhere

Fixes #185


packages/cli/src/cli.ts
```diff
@@ -1,9 +1,9 @@
 #!/usr/bin/env node
 
 import command from './commands'
+import config from './config'
```
andresgutgon, line 4: Before was a singleton but @csansoon changed because he needed to use it on vscode extension. Confirm with him

packages/cli/src/cli.ts
```diff
@@ -69,20 +69,18 @@ CLI.command('credentials')
   .describe('Manage credentials for the data app')
   .option(
     '--create-master-key',
-    'Create a master key for the data app. If you did\'t had one in your .env file',
+    "Create a master key for the data app. If you didn't had one in your .env file",
```
geclos, line 72: Please be careful with user-facing english @latitude-dev/developers

csansoon (approved):

Issue #186, opened 2024-03-25
Provide feedback when installing dependencies in CLI


geclos:
It now looks broken. Anything that makes users wait more than a few seconds should have a visual indicator.

PR #187, opened 2024-03-26
Show in production NPM install progress and also remove pkgManager abstraction

andresgutgon:
# What?
Solve this https://github.com/latitude-dev/latitude/issues/186

### DIFF
--- a/.changeset/great-pianos-decide.md
+++ b/.changeset/great-pianos-decide.md
@@ -1,0 +1,5 @@
+---
+"@latitude-data/cli": minor
+---
+
+Remove pkgManager abstraction and always show npm install progress
--- a/packages/cli/src/lib/setupApp/installDependencies.ts
+++ b/packages/cli/src/lib/setupApp/installDependencies.ts
@@ -5,15 +5,10 @@ import { CLIConfig } from '$src/config'
 export default async function installAppDependencies() {
   const config = CLIConfig.getInstance()
   const appFolder = config.appDir
-  let args = ['install', ...config.pkgManager.flags.mandatoryInstallFlags]
-  args = config.pro
-    ? [...args, config.pkgManager.flags.installFlags.silent]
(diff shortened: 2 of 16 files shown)


packages/cli/src/lib/setupApp/installDependencies.ts
```diff
@@ -5,15 +5,12 @@ import { CLIConfig } from '$src/config'
 export default async function installAppDependencies() {
   const config = CLIConfig.getInstance()
   const appFolder = config.appDir
-  let args = ['install', ...config.pkgManager.flags.mandatoryInstallFlags]
-  args = config.pro
-    ? [...args, config.pkgManager.flags.installFlags.silent]
-    : args
+  let args = ['install', '--legacy-peer-deps']
```
andresgutgon, line 8: This is the real fix. We were checking if it's pro and put `--silent` flag. Now never is silent

geclos: remove the legacy-peer-deps also, I don't think we needed since we fixed the issue with the autoimport package

geclos (approved):

.changeset/great-pianos-decide.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latitude-data/cli": minor
+---
+
+Remove pkgManager abstraction and always show npm install progress
```
geclos, line 5: @andresgutgon this change is gonna require that users install dependencies again if they were previously using pnpm in their project no?

andresgutgon: Probably is only me 😂 

Anyway what should we do?

geclos: change the changeset to indicate that they should do a latitude update in order to reinstall dependencies with npm, if they were previously using pnpm

Issue #177, opened 2024-03-25
Double stdout when prompting for telemetry


geclos:
After choosing an option in the telemetry prompt an identical output gets printed in stdout. Does not affect usage but looks broken.

PR #193, opened 2024-03-26
Avoid showing telemetry UI permissions twice and add tests

andresgutgon:
# What?
Fix this issue https://github.com/latitude-dev/latitude/issues/177

## 🐛 The bug: Double question
image

## Awesome test suite
image

### DIFF
--- a/packages/cli/src/lib/telemetry/index.ts
+++ b/packages/cli/src/lib/telemetry/index.ts
@@ -1,21 +1,20 @@
 import RudderAnalytics from '@rudderstack/rudder-sdk-node'
-import configStore from '../configStore'
 import crypto from 'crypto'
 import os from 'os'
 import { select } from '@inquirer/prompts'
 
 import type { TelemetryEvent, TelemetryEventType } from './events'
 import chalk from 'chalk'
+import configStore from '$src/lib/configStore'
 
 const CLIENT_KEY = '2daExoSEzxW3lPbRFQVoYIGh0Rb'
 const ANALYTICS_URL = 'https://latitudecmggvg.dataplane.rudderstack.com'
-
 type TelemetryConfig = {
   enabled: boolean | undefined
   anonymousUserId: string | undefined
 }
 
-class Telemetry {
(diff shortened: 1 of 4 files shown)


packages/cli/src/lib/telemetry/index.ts
```diff
@@ -41,33 +40,26 @@ class Telemetry {
...
 
   enable() {
-    this.track({ event: 'telemetryEnabled' })
+    this.trackWithoutCheck({ event: 'telemetryEnabled' })
     configStore.set('telemetry.enabled', true)
   }
 
   disable() {
-    this.track({ event: 'telemetryDisabled' })
+    this.trackWithoutCheck({ event: 'telemetryDisabled' })
```
andresgutgon, line 55: This was the bug. We were calling `track` which has `setup` inside so inside a setup we ended calling again `setup` and there for the question appeared twice

packages/cli/src/lib/telemetry/index.ts
```diff
@@ -41,33 +40,26 @@ class Telemetry {
...
   private async setup() {
-    // Only do this one time one running the CLI
     if (this.initialized) return this.enabled
-
-    if (this.enabled !== undefined) {
-      this.initialized = true
-      return this.enabled
-    }
+    if (this.enabled === true) return true
+    if (this.enabled === false) return false
```
csansoon, line 62: `if (this.enabled !== undefined) return this.enabled`

packages/cli/src/lib/telemetry/index.ts
csansoon: Also now that you're with it, there's another bug where sometimes it doesn't add a new line after the last selectable option, so the next message appears right after "No, maybe later\<next message>":

image

csansoon (dismissed):

packages/cli/src/lib/telemetry/index.ts
andresgutgon: This is also fixed with this

Issue #171, opened 2024-03-24
Add sentry


geclos:
We need more visibility on CLI errors. Add optional error reporting via sentry to the CLI.

PR #197, opened 2024-03-27
Add Sentry to CLI

andresgutgon:
# What?
Fix this issue https://github.com/latitude-dev/latitude/issues/171

### TODO
- [x] Add Sentry to CLI
- [x] Configure Sentry with a env variable
- [x] QA it in development with the DSN in my `.env` file
- [x] Configure CI release.yml to build CLI with Sentry enabled. Test and Linter CI task should not enable Sentry.
- [x] Add `CLI_SENTRY_DSN` to github as env variable for building the CLI with Sentry activated. 
- [x] Add `TELEMETRY_URL` and `TELEMETRY_CLIENT_KEY` as github actions secrets
- [x] ~~Think if makes sense to add Sentry to `apps/server` SvelteKit app and how to do it to make it configurable at deploy time~~ Does not makes sense atm
- [x] ~~Configure [Source Maps for CLI for Sentry](https://docs.sentry.io/platforms/node/sourcemaps/)~~ I'll do this in another moment if we see it's necessary

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -38,7 +38,12 @@ jobs:
         uses: changesets/action@v1
         with:
           # This expects us to have a script called release which does a build for our packages and calls changeset publish
-          publish: CLI_SENTRY_ENABLED='true' CLI_SENTRY_DSN=${{ secrets.CLI_SENTRY_DSN }} pnpm release
+          publish: |
+            TELEMETRY_URL=${{ secrets.TELEMETRY_URL }}
+            TELEMETRY_CLIENT_KEY=${{ secrets.TELEMETRY_CLIENT_KEY }}
+            CLI_SENTRY_DSN=${{ secrets.CLI_SENTRY_DSN }}
+            CLI_SENTRY_ENABLED='true'
+            pnpm release
         env:
           GITHUB_TOKEN: ${{ secrets.CUSTOM_GITHUB_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -38,7 +38,7 @@ jobs:
         uses: changesets/action@v1
         with:
           # This expects us to have a script called release which does a build for our packages and calls changeset publish
-          publish: pnpm release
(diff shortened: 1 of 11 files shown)


packages/cli/src/cli.ts
```diff
@@ -10,6 +10,9 @@ import startCommand from './commands/start'
 import telemetryCommand from './commands/telemetry'
 import updateCommand from './commands/update'
 import { onError } from './utils'
+import initSentry from '$src/integrations/sentry'
+
+initSentry()
```
andresgutgon, line 15: I think makes sense to init Sentry here. What do you think?

.github/workflows/release.yml
```diff
@@ -38,7 +38,12 @@ jobs:
         uses: changesets/action@v1
         with:
           # This expects us to have a script called release which does a build for our packages and calls changeset publish
-          publish: pnpm release
+          publish: |
+            TELEMETRY_URL=${{ secrets.TELEMETRY_URL }}
+            TELEMETRY_CLIENT_KEY=${{ secrets.TELEMETRY_CLIENT_KEY }}
+            CLI_SENTRY_DSN=${{ secrets.CLI_SENTRY_DSN }}
```
geclos, line 44: DSN or DNS?

geclos (dismissed):
looks good

.github/workflows/release.yml
andresgutgon: DSN refers to the Data Source Name

packages/cli/rollup.config.mjs
```diff
@@ -1,23 +1,28 @@
...
+  const pkgPath = path.join(DIRNAME, 'package.json')
   return JSON.parse(fs.readFileSync(pkgPath, 'utf8')).version
 }
 
 const nodeEnv = process.env.NODE_ENV ? process.env.NODE_ENV : 'development'
 const packageVersion =
   nodeEnv === 'development' ? 'development-0.0.0' : getPackageVersion()
 
+const workspaceDotEnvPath = resolvePath(DIRNAME, '..', '..', '.env')
+dotenvConfig({ path: workspaceDotEnvPath })
```
andresgutgon, line 24: This is done to use in dev mode when you need to debug these env variables. You put in the root of the project in an `.env` file

PR #198, opened 2024-03-27
Allow to set a debounce time on reactiveToParams option.

csansoon:
Instead of being just a boolean, `reactiveToParams` now allows using a `number` too, to define the debounce time between param updates before calling a refetch.

### DIFF
--- a/apps/server/package.json
+++ b/apps/server/package.json
@@ -30,6 +30,7 @@
     "@testing-library/svelte": "^4.1.0",
     "@types/blessed": "^0.1.25",
     "@types/estree": "^1.0.1",
+    "@types/lodash-es": "^4.17.12",
     "@types/mock-fs": "^4.13.4",
     "autoprefixer": "^10.4.17",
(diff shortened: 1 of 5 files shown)


apps/server/package.json
```diff
@@ -52,8 +53,9 @@
     "@latitude-data/client": "workspace:*",
     "@latitude-data/connector-factory": "workspace:^",
     "@latitude-data/custom_types": "workspace:^",
-    "@latitude-data/svelte": "workspace:*",
+    "@latitude-data/jwt": "workspace:*",
     "@latitude-data/query_service": "workspace:*",
-    "@latitude-data/jwt": "workspace:*"
+    "@latitude-data/svelte": "workspace:*",
+    "lodash": "^4.17.21"
```
andresgutgon, line 59: Use `lodash-es` which is a version optimized for code splitting

csansoon: Ohh I see

andresgutgon (approved):

PR #199, opened 2024-03-27
Fix show template selector when init a project with CLI

andresgutgon:
# What? 
Fix template selector not shown when creating a new project

### TODO
- [x] Fix show template selector when starting a project with CLI
- [x] Show server running URL when showing different version warning

image

image

### DIFF
--- a/packages/cli/src/commands/start/index.ts
+++ b/packages/cli/src/commands/start/index.ts
@@ -34,7 +34,7 @@ async function welcomeMessage() {
 export default async function startCommand({
   name,
   port,
-  template = TemplateUrl.default,
+  template,
   open = false,
 }: {
   open: boolean
@@ -66,6 +66,7 @@ export default async function startCommand({
     template: choosenTemplate,
     force,
   })) as string
+
   config.rootDir = rootDir
 
   await setupApp()
--- a/packages/cli/src/lib/telemetry/index.ts
+++ b/packages/cli/src/lib/telemetry/index.ts
@@ -14,9 +14,13 @@ type TelemetryConfig = {
   anonymousUserId: string | undefined
 }
 
+type TelemetryCredentials = {
+  clientKey: string
+  clientUrl: string
+}
 export class Telemetry {
   private static instance: Telemetry
-  private client: RudderAnalytics
+  private client: RudderAnalytics | undefined = undefined
   private initialized: boolean = false
 
   public static getInstance(): Telemetry {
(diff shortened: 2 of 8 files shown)


packages/cli/src/commands/start/index.ts
```diff
@@ -34,7 +34,7 @@ async function welcomeMessage() {
 export default async function startCommand({
   name,
   port,
-  template = TemplateUrl.default,
```
andresgutgon, line 37: Defaults are the evil 😅

packages/cli/src/lib/telemetry/index.ts
```diff
@@ -130,6 +138,20 @@ export class Telemetry {
   private get config(): TelemetryConfig {
     return configStore.get('telemetry')
   }
+
+  // TODO: Provision TELEMETRY_CLIENT_KEY and TELEMETRY_URL
+  // We don't want to use telemetry in development mode
+  // we will provision this for the published CLI
+  private get credentials(): TelemetryCredentials | undefined {
+    const clientKey = process.env.TELEMETRY_CLIENT_KEY ?? CLIENT_KEY
+    const clientUrl = process.env.TELEMETRY_URL ?? ANALYTICS_URL
```
andresgutgon, line 147: @geclos we are using Telemetry even when we develop in the monorepo. We need to provision these two env variables in GitHub to only enable telemetry when we release the CLI package. I'll do something similar with Sentry

packages/cli/tests/e2e/start.cjs
```diff
@@ -4,7 +4,15 @@ const { spawn, spawnSync } = require('child_process')
...
-spawnSync('node', [path.join(__dirname, '../../dist/cli.js'), 'telemetry', '--disable'])
+// Disable telemetry
+spawnSync(
+  'node',
+  [
+    path.join(__dirname, '../../dist/cli.js'),
+    'telemetry',
+    '--disable'
+  ]
+)
```
andresgutgon, line 15: Telemetry now has a good test coverage. Also with the env variables this won't be necessary. 

## When telemetry should be used
1. When developing the monorepo 🚫 
2. When running CI test or linter 🚫 
3. When publishing the CLI package in NPM ✅

csansoon (dismissed):

packages/cli/src/lib/telemetry/index.ts
geclos: create an issue please 😅 otherwise i won't remember

andresgutgon: [Done](https://github.com/latitude-dev/latitude/issues/207)

PR #206, opened 2024-03-27
Select component

csansoon:
A new `Select` component to add a dropdown with predefined or dynamic options for the user to choose from.


apps/server/src/lib/autoimports/forms/Select/index.svelte
```diff
@@ -0,0 +1,109 @@
...
+  export let opts: $$Props['opts'] = {}
+
+  export let labels: $$Props["labels"] = undefined;
+  export let values: $$Props["values"] = undefined;
+
+  let inputStore = useViewParam(param, value);
+
+  const result = query
+    ? useQuery({ query, inlineParams, opts })
+    : readable({ isLoading: false } as QueryResultState, () => {}); // This is a dummy store with no data, for when there's no query.
```
andresgutgon, line 46: This could be abstracted inside `useQuery`

apps/server/src/lib/autoimports/forms/Select/index.svelte
```diff
@@ -0,0 +1,109 @@
...
+    if (!$result.data) return undefined;
+    // If no user-defined values key is provided, look for a "value" column. Otherwise, use the first column.
+    return $result.data.fields.find((f) => f.name === "value")?.name ?? $result.data.fields[0]?.name;
+  }
+  $: findLabelsKey = () => {
+    if (labels) return labels; // User-defined labels key
+    if (!$result.data) return undefined;
+    // If no user-defined labels key is provided, look for a "label" column. Otherwise, use the value column.
+    return $result.data.fields.find((f) => f.name === "label")?.name ?? findValuesKey();
+  }
```
andresgutgon, line 65: All this logic should be encapsulated and tested. I know is soon in this PR and you're still finding the right solution but I feel this is going to growth in complexity and it should be a tested module used here in the end.

apps/server/src/lib/autoimports/forms/Select/index.svelte
```diff
@@ -0,0 +1,109 @@
...
+
+  
+  $: findValuesKey = () => {
+    if (values) return values; // User-defined values key
+    if (!$result.data) return undefined;
+    // If no user-defined values key is provided, look for a "value" column. Otherwise, use the first column.
+    return $result.data.fields.find((f) => f.name === "value")?.name ?? $result.data.fields[0]?.name;
+  }
+  $: findLabelsKey = () => {
+    if (labels) return labels; // User-defined labels key
```
andresgutgon, line 61: So if user define labels or values query labels and values are ignored?

packages/client/core/src/theme/ui/dialog/content/index.ts
```diff
@@ -0,0 +1,9 @@
+import { cn } from '../../../utils'
+
+export function cssClass({ className }: { className?: string | null }) {
+  return cn("fixed left-[50%] top-[50%] z-50 grid w-full max-w-lg translate-x-[-50%] translate-y-[-50%] gap-4 border bg-background p-6 shadow-lg sm:rounded-lg md:w-full", className)
```
andresgutgon, line 4: `left-1/2` try to use already existing tailwind classes

packages/client/svelte/src/lib/ui/command/command-dialog.svelte
```diff
@@ -0,0 +1,24 @@
...
+	import type { Command as CommandPrimitive } from "cmdk-sv";
+  import { theme } from "@latitude-data/client"
+
+	type $$Props = DialogPrimitive.Props & CommandPrimitive.CommandProps;
+
+	export let open: $$Props["open"] = false;
+	export let value: $$Props["value"] = undefined;
+</script>
+
+<Dialog.Root bind:open {...$$restProps}>
```
andresgutgon, line 14: Abstract this in a Dialog component that use command

```javascript
<Dialog open={open}>
  <Command .../>
</Dialog>
```

packages/client/svelte/src/lib/ui/dialog/dialog-overlay.svelte
```diff
@@ -0,0 +1,21 @@
...
+
+	type $$Props = DialogPrimitive.OverlayProps;
+
+	let className: $$Props["class"] = undefined;
+	export let transition: $$Props["transition"] = fade;
+	export let transitionConfig: $$Props["transitionConfig"] = {
+		duration: 150,
+	};
+	export { className as class };
+</script>
```
andresgutgon, line 14: Funky 👾

apps/server/src/lib/autoimports/forms/Select/index.svelte
csansoon: No, I'd rather have `useQuery` fail when used with undefined. This is just an exceptional case, as `query` is optional.

apps/server/src/lib/autoimports/forms/Select/index.svelte
csansoon: If the user specifies the `labels` and `values` props, I will use those as the keys for the columns.
If not provided, then by default I will use the columns with the keys 'value' and 'label' or, if they don't exist, then I'll just use the value from the first column.

I think that's the most intuitive way to do this: when the user does not specify the column keys, they will either expect to use the 'label' and 'value' columns, or they expect to use the first column. Either way, this system will satisfy the expected behaviour.

packages/client/svelte/src/lib/ui/command/command-dialog.svelte
csansoon: I don't get it.
The `Dialog` component is divided into different subcomponents, because you can define the header, content, footer and different pieces independently.

apps/server/src/lib/autoimports/forms/Select/index.svelte
andresgutgon: Why is good to fail when query is `undefined`? Just to understand

apps/server/src/lib/autoimports/forms/Select/index.svelte
andresgutgon: Ahhh got it. I got it wrong. `labels` and `values` are the name of the columns in the data no?

packages/client/svelte/src/lib/ui/command/command-dialog.svelte
andresgutgon: these classes are internal implementation of dialog. Shouldn't be exposed outside the component `theme.ui.command.dialog.DIALOG_CONTENT_CSS_CLASS}` Of course you can pass you're own `class` but you shouldn't need to define these classes everywhere where you  use a Dialog.

I would do something like that but the default CSS is not exposed. Cleaner

``` javascript
<Dialog open={open}>
  <Dialog.Content>...</Dialog.Content>
</Dialog>
```
☝️ Even this is too much I would prefer a simpler use

In general a dialog is this

``` javascript
<Dialog title='My Title' description='My (optional) description'>
  ...My content. No need to expose `<Dialog.Content />`
</Dialog>
```

apps/server/src/lib/autoimports/forms/Select/index.svelte
csansoon: Yes!

apps/server/src/lib/autoimports/forms/Select/index.svelte
csansoon: `useQuery` is a function that returns reactive content, and it should be used this way. With this in mind, passing `undefined` as the query name does not compute to any actual query, same as when passing a query name that does not exist, so it should fail to warn the user, as it is probably a mistake.

andresgutgon: But why fail? What's the advantage of failing instead of just doing what you're doing here? also this `useQuery` is implemented in `apps/server` so no other users than the app will use it no?

packages/client/svelte/src/lib/ui/dialog/index.ts
```diff
@@ -0,0 +1,34 @@
...
+	//
+	Root as Dialog,
+	Title as DialogTitle,
+	Portal as DialogPortal,
+	Footer as DialogFooter,
+	Header as DialogHeader,
+	Trigger as DialogTrigger,
+	Overlay as DialogOverlay,
+	Content as DialogContent,
+	Description as DialogDescription,
```
andresgutgon, line 33: As commented in other place I would do a simpler dialog exported here as `Dialog` that can be used without any other primitives. In general we would use this simple element and if at some point we need something different we can use the primitives for the different parts

apps/server/src/lib/autoimports/forms/Select/index.svelte
```diff
@@ -0,0 +1,123 @@
...
+  $: allItems = [...userDefinedItems, ...resultItems];
+
+  const dispatch = createEventDispatcher();
+  const onSelect = (selected: unknown) => {
+    setViewParam(param, selected);
+    dispatch("change", selected);
+  };
+</script>
+
+<div class="grid w-full max-w-sm items-center gap-1.5">
```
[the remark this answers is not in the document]
csansoon: The original `Combobox` doesn't have these text parameters, so when implementing these components into other frameworks they wouldn't match! It is just a custom feature for our app, just like with the `Input` and `DatePicker` components!

apps/server/src/lib/autoimports/forms/Select/index.svelte
csansoon: To warn the user, since IMO that's not the desirable usage and may be just a mistake.
Maybe they're using a variable for the query, where it fails when the query does not exist (even if `undefined`), so maybe the variable is not being configured correctly and the user wouldn't notice otherwise.

packages/client/core/src/theme/ui/command/index.ts
```diff
@@ -0,0 +1,25 @@
...
+}
+
+import * as shortcut from './shortcut'
+import * as separator from './separator'
+import * as list from './list'
+import * as item from './item'
+import * as input from './input'
+import * as group from './group'
+import * as empty from './empty'
+import * as dialog from './dialog'
```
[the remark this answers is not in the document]
csansoon: It is using `Dialog` in that implementation, so it is still an independent element. It is done this way because it's customizing `Dialog.Content` with styles changed for this component, instead of just using the default `Dialog` styles and behaviour.

packages/client/svelte/src/lib/ui/dialog/index.ts
csansoon: But `Dialog` here is made to be able to use it in multiple different ways, with custom `Trigger`, `Portal`, `Header`... implementations. This can be seen in the shadcn component documentation, where it is being used as a regular Dialog, where the implementation used on the `Combobox` is not using every component, and is customizing others. This cannot be done if I just export `Dialog.Root` and just put everything inside in `Dialog.Content`.

In addition to this, when implementing the shadcn's component on a different framework, they wouldn't match. That's why I'd prefer the original component in the `svelte` package to just match the imported one from shadcn, and just extract the styles on the `core` in a way where it would always be used in the same way.

packages/client/svelte/package.json
```diff
@@ -56,7 +56,8 @@
     "@latitude-data/custom_types": "workspace:^",
     "@latitude-data/query_result": "workspace:^",
     "@types/lodash-es": "^4.17.12",
-    "bits-ui": "^0.18.1",
+    "bits-ui": "^0.21.1",
```
csansoon, line 59: Also, I updated the `bits-ui` to have access to a new component not available on `0.18`.

apps/server/src/lib/autoimports/forms/Select/index.svelte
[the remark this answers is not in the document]
andresgutgon: But why can't have original `Combobox` `label` and `description`?

packages/client/svelte/src/lib/ui/dialog/index.ts
andresgutgon: No no. I'm not saying not to export all primitives as you're doing. I'm saying to export one default `Dialog` that use those primitives. In general a user of dialog shouldn't need to know about these primitives.

Regarding shadcn compatibility. That's the nice thing about shadcn. Is a starting point. Once created the component on our side we are the owners we don't need to keep compatibility with shadcn. Why would we need to do that?

packages/client/core/src/theme/ui/command/index.ts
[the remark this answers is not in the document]
andresgutgon: But that's a decision on our side. We can decide that our command use normal (our) dialog styles.

packages/client/svelte/src/lib/ui/dialog/index.ts
csansoon: But still, how do you differentiate between the `Dialog.Trigger` and the `Dialog.Content` for example? With the current implementation, it could be something like this:
```svelte
<Dialog.Root>
  <Dialog.Trigger>
    <WarningIcon />
    Edit Profile
  </Dialog.Trigger>
  <Dialog.Content>
    ... <!-- content -->
  </Dialog.Content>
</Dialog.Root>
```

With a single simpler `Dialog` component that's not possible, is it? With a single `Dialog` component with an implementation already done, we could maybe do something like this:
```svelte
<Dialog.Root>
  <Dialog.Trigger>{label}</Dialog.Trigger>
  <Dialog.Content>
    <slot />
  </Dialog.Content>
</Dialog.Root>
```
But it is still not really customizable. And we would still be using the exported primitives on our implementations.

packages/client/core/src/theme/ui/command/index.ts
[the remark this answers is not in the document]
csansoon: The thing is that we might want some styles for the default `Dialog`, for when it is used as an actual Dialog, but this is setting some styles for this implementation of the `Dialog`, where we're removing the padding, along other stuff, making it look something a bit different, while still using its logic.

apps/server/src/lib/autoimports/forms/Select/index.svelte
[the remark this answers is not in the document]
andresgutgon: No but what's the advantage of not having a `Combobox` that you can't pass `label` or `description`?

packages/client/svelte/src/lib/ui/dialog/index.ts
andresgutgon: Ok then

apps/server/src/lib/autoimports/forms/Select/index.svelte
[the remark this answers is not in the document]
csansoon: There's no advantage, but I just don't think it's needed at all.

All of the app's `Input`, `DatePicker`, `RangeDatePicker` and `Select` components have the `label` and `descriptions` props, but none of the original component do. That's something that can easily be added by a user if they're just using the `svelte` package.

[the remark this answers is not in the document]
andresgutgon: But the consistency of having same size label and description is on the hands of the end user. But if we ship our components with these options we ship a more unified interface. You as you considere but imo is better to ship simple components with this decisions baked inside the component. 

```
<Select label='The label' />
```

Is the same reason for shipping a dialog that set `title` and `description` for the end user. Unified experience controlled from the core.

packages/client/svelte/src/lib/ui/scroll-area/scroll-area.svelte
```diff
@@ -0,0 +1,32 @@
+<script lang="ts">
+	import { ScrollArea as ScrollAreaPrimitive } from "bits-ui";
```
andresgutgon, line 2: This primitive is not interesting inside `@latitude-data/svelte`?

packages/client/svelte/src/lib/ui/scroll-area/scroll-area.svelte
```diff
@@ -0,0 +1,32 @@
...
+		scrollbarXClasses?: string;
+		scrollbarYClasses?: string;
+	};
+
+	let className: $$Props["class"] = undefined;
+	export { className as class };
+	export let orientation = "vertical";
+	export let scrollbarXClasses: string = "";
+	export let scrollbarYClasses: string = "";
+</script>
```
andresgutgon, line 17: 👾

andresgutgon (approved):

apps/server/src/lib/autoimports/forms/input/index.svelte
```diff
@@ -35,18 +35,13 @@
   };
 </script>
 
-<div class="grid w-full max-w-sm items-center gap-1.5">
```
andresgutgon, line 38: Lovely!

docs/components/inputs/select.mdx
```diff
@@ -1,4 +1,122 @@
...
+
+### Dynamic options
+
+You can also use a query to load the options dynamically. This can be useful when you have a large number of options, when the options change frequently or when you want to filter the options based on other parameters.
+
+```jsx
+<Select
+	param="user_id"
+	query="users"
+/>
```
andresgutgon, line 47: Weird indentation 👾

csansoon: No, we do actually use tabs on the docs!

PR #211, opened 2024-03-28
Fix: Latitude run watch not working

csansoon:
Latitude run was working by invoking the `query` script in the server app. The invoking command was `npm run query --watch <queryPath> <queryParams>`. This invoked the script with the following arguments: `["--watch", <queryPath>, <queryParams>]` (of course)

For some reason I do not understand, now npm decided not to pass dashed flags to the script, so it now was invoking the script with `[<queryPath>, <queryParams>]`, completely ignoring the `watch` flag.

I have changed the script so that now it expects 3 arguments, instead of 2 arguments and a flag. Not it must be invoked like this: `npm run query <queryPath> <watch:true|false> <queryParams>`

### DIFF
--- a/.changeset/witty-lizards-cough.md
+++ b/.changeset/witty-lizards-cough.md
@@ -1,0 +1,6 @@
+---
+"@latitude-data/cli": patch
+"@latitude-data/server": patch
+---
+
+Fix: CLI run command was ignoring --watch flag.
--- a/apps/server/scripts/run_query/index.ts
+++ b/apps/server/scripts/run_query/index.ts
@@ -1,48 +1,59 @@
 import { QUERIES_DIR } from '../../src/lib/query_service/find_or_compute'
 import findQueryFile from '@latitude-data/query_service'
 import { createConnector } from '@latitude-data/connector-factory'
-import { render, displayResults, displayError } from './result_display'
+import QueryDisplay from './result_display'
 import chokidar from 'chokidar'
 
-const args = process.argv.slice(2)
-if (args.length < 1) {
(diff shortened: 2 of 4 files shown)


csansoon:
> Try running the script like so:
> 
> ```shell
> npm run query -- --watch <queryPath> <queryParams>
> ```

@geclos Already tried, but didn't work. It was still removing the `--watch` flag

andresgutgon:
The CLI has this at the beginning 
``` javascript
#!/usr/bin/env node
```

Can you try to add it to the script. Maybe args are treated different

geclos:
> > Try running the script like so:
> > 
> > ```shell
> > npm run query -- --watch <queryPath> <queryParams>
> > ```
> 
> @geclos Already tried, but didn't work. It was still removing the `--watch` flag

Hmm can we spend some time to try to understand why this is happening? Usually you should be able to pass flags to npm run commands by including the -- before arguments and I want to make sure it's not a bug we've introduced on our end.

csansoon:
I see, I found the issue.

Apparently, running `pnpm run query --watch queryPath queryParams` does NOT skip the watch flag. But running the same command using `npm` instead, does.

What happened is that at first we were spawning the command using `config.pkgManager.command`, but a few versions ago it was changed to be ran with `npm`. That's why it was working before but not now. Also, running it with `--` does seem to work. Either way, I think it is best to just change the API for this script, as this is not meant to be ran directly by the user.

Also, I have modified the `results_display.ts` module, so that when imported it is not executed right away, as it was interfering with my debugging. Not it will only be instantiated when used instead.

geclos (approved):

Issue #212, opened 2024-03-29
Support ealier vs code versions


geclos:
Our extension currently requires the latest vs code version to be installed (1.87 from February 2024) but some vscode forks are a bit behind (notably Cursor, which requires 1.86 from January 2024), we have to support these earlier versions.

PR #213, opened 2024-03-29
feature: relax vscode version requirements a bit

geclos:
We are requiring the latest vscode version (1.87), but some vscode forks are a bit behind. Relaxing our requirements a bit to add support for those forks.

Related #212


andresgutgon:
No changelog

geclos:
hmm there's no changelog for the extension, we will have to automate the release process at some point, right now it's very manual

PR #222, opened 2024-04-02
feature: add pre step to verify we are in a latitude project

geclos:
CLI is currently creating a latitude.json in every folder it runs which is not the best experience. We should ensure our templates have a latitude.json file and fail whenever the CLI runs in an environment that is not a Latitude project.


packages/cli/src/lib/decorators/isLatitudeProject/index.ts
```diff
@@ -0,0 +1,19 @@
...
+        message:
+          'The current directory is not a Latitude project. Run `latitude setup` to set it up as a Latitude project.',
+      })
+
+      process.exit(1)
+    }
+
+    return commandFn(...args)
+  }
+}
```
csansoon, line 19: Maybe rename it to `assertIsLatitudeProject`, `enforceLatitudeProject`, `mustBeLatitudeProject` or something like that. `isLatitudeProject` sounds like it just returns a boolean.

packages/cli/src/lib/decorators/isLatitudeProject/index.ts
```diff
@@ -0,0 +1,19 @@
...
+import { onError } from '$src/utils'
+import fs from 'fs'
+import path from 'path'
+
+export default function isLatitudeProject(commandFn: Function) {
+  return async function (...args: any[]) {
+    if (!fs.existsSync(path.join(process.cwd(), LATITUDE_CONFIG_FILE))) {
+      onError({
+        message:
+          'The current directory is not a Latitude project. Run `latitude setup` to set it up as a Latitude project.',
```
csansoon, line 11: "... or move to an actual Latitude project" (?)

andresgutgon (dismissed):

PR #223, opened 2024-04-02
feature: run CLI commands from any subdirectory

geclos:
Allow users to run CLI commands from any subdirectory


packages/cli/src/lib/decorators/setRootDir/index.ts
```diff
@@ -0,0 +1,47 @@
...
+import { onError } from '$src/utils'
+
+export function findRootDir(maxTries = 100) {
+  let currentDir = process.cwd()
+  let tries = 0
+
+  /* eslint-disable no-constant-condition */
+  while (true) {
+    const latitudeJsonPath = path.join(currentDir, LATITUDE_CONFIG_FILE)
+    if (tries >= maxTries) {
```
andresgutgon, line 14: Where's `maxTries` defined?

andresgutgon (approved):

Issue #105, opened 2024-03-11
React components


geclos:
# WAT
We want to offer a standalone library of react components that communicate with a Latitude backend, so that you can easily integrate more deeply Latitude into existing codebases.

## TODO
- [x] Implement data layer in `@latitude-data/react` [here](https://github.com/latitude-dev/latitude/pull/225/files)
- [x] Implement `latitude-embed` JS component to communicate with an embedded iframe of a latitude app [here](https://github.com/latitude-dev/latitude/pull/286)

Thawab8:
can this enable us to build our own component? for example using this as date range https://ui.shadcn.com/docs/components/date-picker

geclos:
Yup, as part of the react library there will be a state manager to handle communication with a latitude backend, so that you can integrate create your own components that communicate with Latitude.

What I'm thinking about (pseudo-code, don't take it too seriously):

```typescript
import { useQuery } from '@latitude-data/react'

function MyCustomReactComponent () {
  const { data: queryResult, isLoading } = useQuery('titles') // this takes care of communicating with a Latitude backend and holding the frontend state

  console.log(queryResult) // or do whatever you want with it

  // ...
}
```

andresgutgon:
@Thawab8 documentation for getting data from your Latitude project in your react project can be found here 
https://docs.latitude.so/guides/embed/react-components

Please let us know if something is not clear

PR #225, opened 2024-04-03
React package with initial latitude store implementation

andresgutgon:
# What?
We want to start offering a react implementation of latitude data store to get queries from our Latitude server

## Issue 
https://github.com/latitude-dev/latitude/issues/105

### TODO
- [x] Implement react latitude store with `@tanstack/react`
- [x] Add test for `compute` and check `__force=true` is sent to the API
- [x] Make an example using a latitude app server + a simple Vite-React app using `useLatitudeQuery`
- [x] Add docs on how to use 
- [x] QA `apps/server` keeps working. I did a small change in core store in the way we use our api lib


packages/client/core/src/stores/queries.ts
```diff
@@ -85,32 +84,20 @@ export const store = createStore<StoreState>((set, get) => {
...
     },
     forceRefetch: async ({ queryPath, params }) => {
       const queryKey = createQueryKey(queryPath, params || {})
-      performQueryFetch(queryKey, async () => {
-        const response = await api.get<QueryResultPayload>(
-          `api/queries/${queryPath}`,
-          { ...params, [FORCE_REFETCH_PARAMETER]: 'true' }
-        )
-
-        return new QueryResult(response)
```
andresgutgon, line 104: Returning a `QueryResult` here was redundant. Please confirm @csansoon

packages/client/core/src/data/api.ts
```diff
@@ -1,12 +1,14 @@
 import { format } from '@latitude-data/custom_types'
-import { FORCE_REFETCH_PARAMETER } from '../stores/queries';
```
andresgutgon, line 2: We had a circular dependency here. `queries` use api and api use this constant from queries. I introduced it some PRs ago

packages/client/react/src/data/LatitudeProvider.tsx
```diff
@@ -0,0 +1,61 @@
...
+      queries: {
+        enabled: true,
+        refetchOnWindowFocus: false,
+        refetchInterval: false,
+        refetchIntervalInBackground: false,
+        staleTime: Infinity,
+        refetchOnReconnect: 'always',
+        refetchOnMount: false,
+        notifyOnChangeProps: 'all',
+        placeholderData: EMPTY_PAYLOAD,
```
andresgutgon, line 48: I had the idea of adding a default initial state. `placeholderData` doesn't count as real data for this query key but this way users can rely on the basic shape of a latitude query to be there. I'll check if this has some drawbacks.

packages/client/react/src/data/LatitudeProvider.tsx
```diff
@@ -0,0 +1,61 @@
...
+        ...queriesConfig,
+      },
+    },
+  })
+
+  return (
+    <QueryClientProvider client={queryClient}>
+      <LatitudeContext.Provider value={{ api: new LatitudeApi(apiConfig) }}>
```
andresgutgon, line 56: Allow users to set `cors` and `host` before start using the latitude API

packages/client/svelte/vite.config.ts
```diff
@@ -1,12 +1,6 @@
 import { sveltekit } from '@sveltejs/kit/vite'
-import { defineConfig } from 'vitest/config'
+import { defineConfig } from 'vite'
```
andresgutgon, line 2: I don't understand why this didn't fail before
image

packages/client/react/src/data/LatitudeProvider.tsx
```diff
@@ -0,0 +1,60 @@
+import { JSX, useContext, ReactNode, createContext } from 'react'
+import { QueryClient, QueryClientProvider, type QueryClientConfig } from '@tanstack/react-query'
+import { LatitudeApi } from '@latitude-data/client'
+
+type ILatitude = { api: LatitudeApi }
+const LatitudeContext = createContext<ILatitude>({ api: new LatitudeApi() })
+
+import { LatitudeApiConfig } from '@latitude-data/client'
+export type QueryClientProviderProps = {
```
geclos, line 9: 👾

packages/client/react/src/data/useQuery.ts
```diff
@@ -0,0 +1,47 @@
...
+    setComputing(true)
+    const data = await api.getQuery({ queryPath, params, force: true })
+    queryClient.setQueryData(queryKeyProp, data)
+    setComputing(false)
+  }, [queryClient, queryKeyProp])
+
+  return {
+    ...query,
+    compute,
+    isFetching: query.isFetching || isComputing || query.isPlaceholderData,
```
geclos, line 45: the last check worries me why are we fetching if isPlaceholderData returns true? shouldn't the isfetching+isComputing check be enough?

packages/client/react/src/data/useQuery.ts
```diff
@@ -0,0 +1,47 @@
...
+  const query = useReactQuery({
+    queryKey: queryKeyProp,
+    queryFn: async () => api.getQuery({ queryPath, params, force: false }),
+    ...(tanstaskQueryOptions ?? {}),
+    enabled: tanstaskQueryOptions?.enabled ?? true,
+  })
+
+  const compute = useCallback(async () => {
+    setComputing(true)
+    const data = await api.getQuery({ queryPath, params, force: true })
```
geclos, line 37: can this hard crash, what happens if it does

packages/client/react/src/data/useQuery.ts
```diff
@@ -0,0 +1,47 @@
...
+  const queryClient = useQueryClient()
+  const { api } = useLatitude()
+  const queryKeyProp = useMemo(
+    () => [createQueryKey(queryPath, params)],
+    [queryPath, params],
+  )
+  const query = useReactQuery({
+    queryKey: queryKeyProp,
+    queryFn: async () => api.getQuery({ queryPath, params, force: false }),
+    ...(tanstaskQueryOptions ?? {}),
```
geclos, line 31: add a default param instead [here](https://github.com/latitude-dev/latitude/pull/225/files#diff-7febede2261920672d1e5ff6b70cdb0e7e763a06d38adb187ed1fdca97f80361R19)

packages/client/react/src/data/useQuery.ts
```diff
@@ -0,0 +1,47 @@
...
+
+type Props = {
+  queryPath: string
+  params?: QueryParams
+  tanstaskQueryOptions?: Exclude<UseQueryOptions<QueryResultPayload>, 'queryKey' | 'queryFn'>
+}
+export function useQuery({
+  queryPath,
+  params = {},
+  tanstaskQueryOptions,
```
geclos, line 19: default to `{ enabled: true }` if nothing is passed

packages/client/react/src/data/useQuery.ts
andresgutgon: `isPlaceholderData` is only when there is no data in the react query cache. With placeholder data now what we have always from start is 
```javascript
{
  fields: [],
  rows: [],
  rowCount: 0
}
```

React query knows when the cache has something or has that placeholder. And placeholder is true only before fetching. We can easily remove this placeholder thing if we see it doesn't work as expected. But in the example I did and in the tests looks ok for now

pnpm-workspace.yaml
```diff
@@ -2,3 +2,4 @@ packages:
   - 'apps/*'
   - 'packages/**'
   - 'tools/**'
+  - 'examples/example-react'
```
geclos, line 5: why this? i wouldn't treat the examples as a signifcant part of the monorepo

packages/client/react/src/data/useQuery.ts
andresgutgon: Ah ok then

pnpm-workspace.yaml
andresgutgon: Combinient for installed the react package before even publishing it. Otherwise I have to do mambo jambo with `npm link`

packages/client/react/src/data/useQuery.ts
andresgutgon: Hard crash you mean a 500 in the server right?

packages/client/react/src/data/useQuery.ts
geclos: yes but that there is placeholder data it does not mean we are fetching no? and if it does mean that, isn't it already covered by the query.isFetching boolean? I'm ok with having placeholder data, what i don't get is its role in the `isFetching` boolean

pnpm-workspace.yaml
geclos: but why the react example is not related to the rest of the monorepo in any way no? i can just go to its folder and work in it as if it were a separate repo

andresgutgon: Is related in the sense that use a package of the monorepo. What's the problem you see by placing examples as a monorepo folder?

packages/client/react/src/data/useQuery.ts
andresgutgon: ok I'll remove

examples/sample-react/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "version": "0.0.0",
+  "type": "module",
+  "scripts": {
+    "dev": "vite",
+    "build": "tsc && vite build",
+    "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
+    "preview": "vite preview"
+  },
+  "dependencies": {
+    "@latitude-data/react": "workspace:*",
```
geclos, line 13: oh i see, nein. Let's not link the monorepo these examples should be of real usage, if someone copy pastes this example to their machine (not the whole monorepo) nothing will work

pnpm-workspace.yaml
geclos: oh [i see now](https://github.com/latitude-dev/latitude/pull/225#discussion_r1551884190)

examples/sample-react/package.json
andresgutgon: I'll merge this way and then put the real version. Also remove it from workspaces

packages/client/react/src/data/useQuery.ts
geclos: `await` will throw if that promise rejects, not sure when the fetch rejects but i know it does

andresgutgon: Now is try/catched

apps/server/src/hooks.server.ts
```diff
@@ -0,0 +1,40 @@
...
+    name: 'Access-Control-Allow-Methods',
+    value: 'GET, OPTIONS',
+  },
+  headers: {
+    name: 'Access-Control-Allow-Headers',
+    value: 'Content-Type, Accept',
+  },
+}
+export const handle: Handle = async ({ event, resolve }) => {
+  const response = await resolve(event)
```
geclos, line 21: another unchaught await, can it hard crash?

examples/sample-react/package.json
andresgutgon: Moving to to sites mean not pushing in this PR. Why is not ok doing this way if I'm going to change right away after publish?

geclos: it's fine it's fine

geclos (approved):

PR #230, opened 2024-04-03
iframe example added to components and visualizations

samulatitude:


docs/components/interaction/run-button.mdx
```diff
@@ -7,6 +7,8 @@ description: 'Add buttons to your views to update the data in your app'
 
 The `RunButton` component lets users run a query—or a group of queries—by clicking a button.
 
+<iframe class="rounded-xl" src="https://docs-charts-examples.fly.dev/components/run-button" width="100%" height="120"> </iframe>
```
andresgutgon, line 10: I'm a bit worry that we depend now on fly to be up for our docs to look good. 

I think maybe we can explore deploying a static version of our Storybook that will never be down.

samulatitude: I think fly is ok for the moment, super easy to deploy and it's unlikely they go down no?

andresgutgon: Yeah, not sure we're in the free tier so the readability is not like pro : )

Anyway is just a comment, let's go with this

andresgutgon (approved):

PR #234, opened 2024-04-04
Fix: remove max width for input components

csansoon:


.changeset/rare-spies-explain.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latitude-data/client": minor
+---
+
+Remove max width in input components
```
csansoon, line 5: Should it be a `minor` or just a `patch`?

geclos: patch is fine

geclos (approved):

PR #238, opened 2024-04-05
Add react example app to the project

andresgutgon:
## What?
Example react app using `@latitude-data/react@0.1.0`

image


examples/sample-react/.gitignore
```diff
@@ -0,0 +1,29 @@
...
+*.suo
+*.ntvs*
+*.njsproj
+*.sln
+*.sw?
+
+# Do not commit lock files in the example
+package-lock.json
+yarn.lock
+pnpm-lock.yaml
```
andresgutgon, line 29: I think this is better also if you use `pnpm` install independently from monorepo 

```bash
pnpm install --ignore-workspace
```

PR #239, opened 2024-04-05
fix: Select combobox alignment

geclos:
Left align popover with it's trigger. Expose the prop.


packages/client/svelte/src/lib/ui/combobox/index.svelte
```diff
@@ -79,10 +81,13 @@
         <CaretSort class={theme.ui.combobox.BUTTON_ICON_CSS_CLASS} />
       </Button>
     </Popover.Trigger>
-    <Popover.Content class={theme.ui.combobox.POPOVER_CSS_CLASS}>
+    <Popover.Content {align} class={theme.ui.combobox.POPOVER_CSS_CLASS}>
```
geclos, line 84: the only real thing in this PR is this

andresgutgon (approved):

PR #241, opened 2024-04-05
Introduce WebComponents packages managed with StencilJS

andresgutgon:
# What?
We want to allow users to embed a latitude data app view inside their React or Vue or VanillaJS projects and interact with the params in the latitude view or pass the signup token or listen to params changes inside the view. 

Technically I think using Web Component is a good solution because it allows us to add logic inside the component. I want to try [StencilJS](https://stenciljs.com/) for building the primitive and then doing also the react wrapper. 

According to their docs they have integrations with all mayor frameworks so we can add Svelte, Vue or Angular in the future.

## TODO
- [x] Introduce `@latitude-data/webcomponents` to have WebComponents
- [x] Add a basic simple example as foundation for `<latitude-embed>`. Our first component.
- [x] Implement React wrapper around the web component primitive of `<latitude-embed>`

### Develop LatitudeEmbed component (another PR)
- [ ] Allow to pass and update `signedToken` and update internal iframe URL
- [ ] Allow host to listen to params changes inside the iframe
- [ ] Allow host to update one or more params and the iframe's URL to be changed. 
- [ ] Allow host to run all or some queries present in the latitude view
- [ ] `latitude-embed` has to be responsive. When the host size change inner iframe has to be updated.
- [ ] Add documentation


examples/sample-react/package.json
```diff
@@ -7,12 +7,14 @@
     "dev": "vite",
     "build": "tsc && vite build",
     "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
+    "prettier": "prettier --write 'src/*.ts' 'src/**/*.tsx' 'src/**/*.ts'",
     "preview": "vite preview"
   },
   "dependencies": {
-    "@latitude-data/react": "0.1.0",
+    "@latitude-data/react": "workspace:*",
```
andresgutgon, line 14: ⚠️ Revert this change

packages/client/react/tsconfig.json
```diff
@@ -11,7 +11,7 @@
       "$src/*": ["src/*"]
     },
     "esModuleInterop": true,
-    "skipLibCheck": true,
+    "strictNullChecks": false,
```
andresgutgon, line 14: I had to add this flag until this issue is solved
https://github.com/ionic-team/stencil-ds-output-targets/issues/404#issuecomment-2039982714

PR #246, opened 2024-04-05
Feature/csv export button in charts

geclos:
- Adds csv export popover to all charts and tables
- Adds `download` to React's useQuery

### DIFF
--- a/apps/server/src/lib/autoimports/chart/FunnelChart.svelte
+++ b/apps/server/src/lib/autoimports/chart/FunnelChart.svelte
@@ -2,19 +2,32 @@
   import type { FunnelChartProps } from '@latitude-data/svelte/internal'
   import { useQuery, type QueryProps } from '$lib/stores/queries'
   import { FunnelChart } from '@latitude-data/svelte/internal'
-  type $$Props = Omit<FunnelChartProps, 'data' | 'isLoading' | 'error'> &
-    QueryProps
   import { DEFAULT_HEIGHT } from './constants'
+  import { api } from '@latitude-data/client'
+
+  type $$Props = Omit<
+    FunnelChartProps,
+    'download' | 'data' | 'isLoading' | 'error'
+  > &
+    QueryProps & {
+      download?: boolean
+    }
 
   export let query: $$Props['query']
   export let inlineParams: $$Props['inlineParams'] = {}
   export let opts: $$Props['opts'] = {}
   export let height: $$Props['height'] = DEFAULT_HEIGHT
+  export let download: $$Props['download'] = false
+
+  const downloadFn = download
+    ? () => api.downloadQuery({ queryPath: query, params: inlineParams })
+    : undefined
 
   const result = useQuery({ query, inlineParams, opts })
 </script>
 
 <FunnelChart
+  download={downloadFn}
   {height}
   data={$result.data}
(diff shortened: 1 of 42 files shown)


packages/client/core/src/data/api.ts
```diff
@@ -50,16 +54,28 @@ export class LatitudeApi {
...
+  }
+
+  async downloadQuery({
+    queryPath,
+    params = {},
+  }: {
+    queryPath: string
+    params?: QueryParams
+  }) {
+    return this.get<Blob>(`api/queries/${queryPath}`, {
```
andresgutgon, line 69: Dry use `getQuery<Blob>` just make `getQuery` generic and make `download` a flag

```javascript
return this.getQuery<Blob>({ queryPath, params, force, download: true })
```

```javascript
getQuery<T>: Promise<T>
```

andresgutgon: Because if not download can't be forced

packages/client/core/src/theme/ui/card/index.ts
```diff
@@ -18,8 +17,8 @@ const cardHeaderVariants = tv({
   base: 'flex flex-col space-y-1.5',
   variants: {
     type: {
-      normal: NORMAL_PADDING,
-      invisible: 'pb-6',
+      normal: 'pb-6',
+      invisible: '',
```
andresgutgon, line 21: How looks this now?

packages/client/svelte/src/lib/internal/ui/chart/_Wrapper.svelte
```diff
@@ -99,10 +103,45 @@
...
+                      <circle cx="8" cy="12" r="1" fill="currentColor" />
+                      <circle cx="12" cy="12" r="1" fill="currentColor" />
+                      <circle cx="16" cy="12" r="1" fill="currentColor" />
+                    </svg>
+                  </div>
+                </Popover.Trigger>
+                <Popover.Content class="flex px-0 py-2">
+                  <div class="w-full px-2 hover:bg-gray-100">
+                    <button on:click={download}>
+                      <Text size="h5">Download as CSV</Text>
```
andresgutgon, line 136: Yep i18n is a thing : /

packages/client/core/src/theme/ui/card/index.ts
geclos: i've removed some extra padding we do not want

apps/server/src/lib/autoimports/chart/FunnelChart.svelte
```diff
@@ -2,19 +2,32 @@
...
   import { DEFAULT_HEIGHT } from './constants'
+  import { api } from '@latitude-data/client'
+
+  type $$Props = Omit<
+    FunnelChartProps,
+    'download' | 'data' | 'isLoading' | 'error'
+  > &
+    QueryProps & {
+      download?: boolean
+    }
```
andresgutgon, line 14: This is repeated everywhere. We could have a generic:

``` javascript
type VisualizationProps<T extends ChartProps> = Omit<T, 'download' | 'data' | 'isLoading' | 'error'> & {
  download?: boolean
}
```

andresgutgon (dismissed):

PR #259, opened 2024-04-08
Fix: Manage connection lifespan

csansoon:
`BaseConnector` is already managing the connector pools lifespan. It runs the `connect` method before running a query, then runs that query and any other one that is included in it, and finally runs the `disconnect` method afterwards.

This allows to keep the `BaseConnector` instance alive whenever we want, without having to worry about having open connections.

Although this is the default behaviour, most connectors do not have any implementation for these `connect` and `disconnect` methods, which they should.

### DIFF
--- a/apps/server/src/lib/server/connectorManager.ts
+++ b/apps/server/src/lib/server/connectorManager.ts
@@ -1,0 +1,21 @@
+import { type BaseConnector } from '@latitude-data/base-connector'
+import { createConnector } from '@latitude-data/connector-factory'
+
+const connectors: Record<string, BaseConnector> = {}
+
+export function loadConnector(sourcePath: string): BaseConnector {
+  if (!connectors[sourcePath]) {
+    connectors[sourcePath] = createConnector(sourcePath)
+  }
+
+  return connectors[sourcePath]
+}
+
+export async function clearConnector(sourcePath: string): Promise<void> {
+  await connectors[sourcePath]?.end()
+  delete connectors[sourcePath]
+}
+
+export async function clearAllConnector(): Promise<void> {
+  await Promise.all(Object.keys(connectors).map(clearConnector))
+}
--- a/apps/server/src/lib/server/shutdown.ts
+++ b/apps/server/src/lib/server/shutdown.ts
@@ -1,0 +1,36 @@
+import { clearAllConnector } from './connectorManager'
+let isListening = false
+
+/**
+ * This module provides mechanisms to gracefully shutdown of the server. A graceful shutdown ensures that the
+ * server terminates its operations in an orderly manner, handling any necessary cleanup tasks before exiting.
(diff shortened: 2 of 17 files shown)


packages/connectors/mssql/src/index.ts
```diff
@@ -23,14 +23,18 @@ export type ConnectionParams = {
...
-  private pool
+  private pool: sql.ConnectionPool
 
   constructor(rootPath: string, connectionParams: ConnectionParams) {
     super(rootPath)
 
     this.pool = new sql.ConnectionPool(connectionParams)
   }
 
+  end(): Promise<void> {
```
andresgutgon, line 34: Looking in general the majority of connectors call this `close`. I would call this method close too

andresgutgon (approved):

packages/connectors/duckdb/src/index.ts
```diff
@@ -12,28 +12,37 @@ export type ConnectionParams = {
...
     }
   }
 
+  private async createClient(): Promise<void> {
+    this.client = await Database.create(
+      this.url,
+      this.url === ':memory:' ? OPEN_READWRITE : OPEN_READONLY,
+    )
+  }
+
```
csansoon, line 41: Duckdb is the only connector that does not create the Client in the class constructor, but instead it is created on the first `runQuery`. This is because the function that creates the query is asyncronous, and thus cannot be used in the `constructor` function.

You could do something like `Database.create(...).then((client) => { this.client = client })`, but typescript doesn't like that, since it is not certain that the variable will be created right after the constructor. Instantiating the connector and running `runQuery` right afterwards could result in an error if it takes some time to create the client.

apps/server/src/lib/server/shutdown.ts
```diff
@@ -0,0 +1,36 @@
+import { clearAllConnector } from './connectorManager'
```
geclos, line 1: clearAllConnectors that missing s is killing me

apps/server/src/lib/server/connectorManager.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  return connectors[sourcePath]
+}
+
+export async function clearConnector(sourcePath: string): Promise<void> {
+  await connectors[sourcePath]?.end()
+  delete connectors[sourcePath]
+}
+
+export async function clearAllConnector(): Promise<void> {
```
geclos, line 19: that missing trailing s please, my eyes are bleeding

packages/connectors/mysql/src/index.ts
```diff
@@ -58,8 +66,6 @@ export class MysqlConnector extends BaseConnector {
           (error, results, fields) => {
             connection.release()
 
-            this.pool.end()
```
geclos, line 61: thanks 🙏🏼

geclos (commented):
lovely

PR #261, opened 2024-04-09
Fix: Close server gracefully on SIGINT and SIGTERM

csansoon:


apps/server/package.json
```diff
@@ -6,7 +6,9 @@
   "type": "module",
   "scripts": {
     "dev": "vite dev",
-    "build": "svelte-kit sync && vite build",
+    "build:server": "svelte-kit sync && vite build",
+    "build:shutdownPlugin": "mv build/index.js build/server.js && cp plugins/shutdownOnDemand.js build/index.js",
```
andresgutgon, line 10: Why you call it plugins?

apps/server/plugins/shutdownOnDemand.js
```diff
@@ -0,0 +1,13 @@
...
+  server.server?.close(() => {
+    process.exit(0);
+  });
+  server.server?.closeIdleConnections();
+  setInterval(() => server.server?.closeIdleConnections(), 1_000);
+  setTimeout(() => server.server?.closeAllConnections(), 20_000);
+}
+
+process.on('SIGINT', shutdownServer);
+process.on('SIGTERM', shutdownServer);
```
andresgutgon, line 13: So this file is not responsible of starting the Svelte server right?

apps/server/package.json
csansoon: Yeah... Should I move it to `scripts`, maybe?

csansoon: Moved it to `scripts`

geclos (approved):
coolio

andresgutgon (approved):

PR #263, opened 2024-04-09
Feature/auth latitude server

geclos:
Implements authentication against the Latitude server. Will be used for our paid infrastructure service.

### DIFF
--- a/packages/cli/src/commands/logout/index.ts
+++ b/packages/cli/src/commands/logout/index.ts
@@ -1,0 +1,7 @@
+import configStore from '$src/lib/configStore'
+
+export default function logoutCommand() {
+  configStore.set('jwt', undefined)
+
+  console.log('Logout successful.')
+}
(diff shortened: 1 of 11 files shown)


packages/cli/src/commands/logout/index.ts
```diff
@@ -0,0 +1,7 @@
+import configStore from '$src/lib/configStore'
+
+export default function logoutCommand() {
+  configStore.set('jwt', undefined)
```
andresgutgon, line 4: `jwt` is not clear. I would call it `currentUser` or something like that

packages/cli/src/commands/signup/index.ts
```diff
@@ -0,0 +1,49 @@
...
+      console.error('Signup request failed:', err.message)
+    } else {
+      const { response, responseBody } = res
+      if (response.statusCode === 200) {
+        try {
+          const response = JSON.parse(responseBody)
+          const jwt = response.token
+          if (jwt) {
+            configStore.set('jwt', jwt)
+            console.log(chalk.green('Signup successful! JWT stored.'))
```
andresgutgon, line 32: Better message. `Signup successful! with email@gmail.com`

andresgutgon (approved):

PR #266, opened 2024-04-09
Feature: reload queries on hot reload

csansoon:
Refetch a query on the client when the sql file has been updated. This refetch is done with the `__force` flag, to make sure the updated data is shown.


andresgutgon (dismissed):
really nice

PR #273, opened 2024-04-11
View docs reorganization

csansoon:
Reorganised and rewrite some documentation pages to be more easy to follow.

Changes:
 - **Queries** section now only has information about query building and execution. **Parameters** section inside no longer explains how to configure the parameters in the view.
 - New **Global Parameters** section in **Views** explaining what global parameters are and how they work.
 - New **Running queries** section in **Views** explaining how to request queries and configure them from the View.
 - New **{param(...)}** section in **Views/Logic** explaining how to use and display values from parameters in the view.
 - **{#await ...}** now explains how to use **runQuery(...)** in greater detail
 - Moved **Components** and **Visualizations** tabs into **Views** inside a **Components** section, which then is organised with sub-categories.
 - Reorganized the documentation on Visualization properties to be better categorized.
 - Removed and fixed references to some unexisting functions like `url_param(...)`


geclos (approved):

docs/guides/development/latitude-cli.mdx
```diff
@@ -59,10 +59,10 @@ With this, every time you save the file, the query will be re-run and the result
 
 ### Passing parameters
 
-If you're using parameters in your query, you can pass them to the `run` command like this:
+If you're using parameters in your query, you can pass them to the `run` command using the `--param` flag, followd by the parameter name and value. For example:
```
cesr, line 62:
```suggestion
If you're using parameters in your query, you can pass them to the `run` command using the `--param` flag, followed by the parameter name and value. For example:
```

docs/queries/logic/parameters.mdx
```diff
@@ -1,256 +1,40 @@
...
+In the query, the parameters can be used with the `param()` function. This function retrieves the value specified by the user and inserts it safely into the query.
 
-``` sql
-sql my_query.sql
+```sql
 SELECT *
-FROM table
-WHERE id != {param('company_id')} -- This is where the data from Input is inserted.
+FROM companies
+WHERE id == {param('company_id')}
```
cesr, line 19: is the double == valid sql?

cesr (commented):
Absolutely LOVE this revamp, great work @csansoon!

docs/queries/logic/parameters.mdx
csansoon: Actually I think it's not 👌

PR #281, opened 2024-04-12
Static files support for views

csansoon:
ALL files inside `views` that are not `html` files will be copied to the `static` directory in the app folder. The only exception are hidden files and directories (starting with a dot).

We were using a `static/latitude` folder for the queries, but since it can now conflict with the user's view files, I changed it to `static/.latitude`, since hidden directories are not copied.

### DIFF
--- a/packages/cli/src/lib/sync/syncQueries/index.ts
+++ b/packages/cli/src/lib/sync/syncQueries/index.ts
@@ -25,7 +25,7 @@ export default async function syncQueries({
     rootDir,
     APP_FOLDER,
     'static',
-    'latitude',
+    '.latitude',
     'queries',
   )
 
--- a/packages/cli/src/lib/sync/syncStaticFiles/index.ts
+++ b/packages/cli/src/lib/sync/syncStaticFiles/index.ts
@@ -6,27 +6,27 @@ import watcher from '../shared/watcher'
 import { onExit } from '$src/utils'
 import config from '$src/config'
 
-function getRoutesFolderPath(cwd: string): string {
-  return path.join(cwd, APP_FOLDER, 'src', 'routes')
+function getStaticFilesFolderPath(cwd: string): string {
+  return path.join(cwd, APP_FOLDER, 'static')
 }
+
+const IGNORED_REGEX = /^.*(?:\/|^)(\..+|.+\.html)$/ // Ignore all html files and hidden files and directories (starting with a dot)
 
 const copiedFiles = new Set<string>()
 
-const IGNORED_FILES_REGEX = /^(?!.*index\.html$).*$/ // ignore all files except index.html
-
-export default async function syncViews({
+export default async function syncStaticFiles({
   watch = false,
 }: {
   watch?: boolean
 }): Promise<void> {
   const rootDir = config.rootDir
-  const destinationDir = getRoutesFolderPath(rootDir)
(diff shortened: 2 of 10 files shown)


packages/cli/src/lib/sync/syncViews/index.ts
```diff
@@ -24,7 +26,7 @@ export default async function syncViews({
 
   if (watch) {
     await watcher(viewsDir, syncFn, {
-      ignored: /(?!.*\/index\.html$)(^|[/\\])\../, // ignore all files except index.html
```
csansoon, line 27: This regex didn't work as expected. It was copying other files like images too.

packages/cli/src/lib/sync/syncViews/index.ts
```diff
@@ -24,7 +26,7 @@ export default async function syncViews({
...
+      // ignored: IGNORED_FILES_REGEX,
     })
```
csansoon, line 30: This regex didn't work as expected. It was copying other files like images too.

packages/cli/src/lib/sync/syncQueries/index.ts
```diff
@@ -25,7 +25,7 @@ export default async function syncQueries({
     rootDir,
     APP_FOLDER,
     'static',
-    'latitude',
+    '.latitude',
```
andresgutgon, line 28: Why is `.` now?

csansoon: All files inside `views` that are not `index.html` will be copied to the `static` directory in the app folder. The only exception are hidden files and directories (starting with a dot).

We were using a `static/latitude` folder for the queries, but since it can now conflict with the user's view files, I changed it to `static/.latitude`, since hidden directories are not copied.

packages/cli/src/lib/sync/syncStaticFiles/index.ts
```diff
@@ -0,0 +1,90 @@
...
+import { APP_FOLDER } from '$src/commands/constants'
+import watcher from '../shared/watcher'
+import { onExit } from '$src/utils'
+import config from '$src/config'
+
+function getStaticFilesFolderPath(cwd: string): string {
+  return path.join(cwd, APP_FOLDER, 'static')
+}
+
+const IGNORED_REGEX = /^.*(?:\/|^)(\..+|index\.html)$/ // Ignore all index.html files and hidden files and directories (starting with a dot)
```
andresgutgon, line 13: Ah you do the `.` for this no?

What about doing the other way. Do a simple white list of image extensions with the more common and only copy those `.png`, `.jpg`, `.jpeg`, `.gif`,...some more modern extensions and that's it

csansoon: I think it's better to just copy any file, so we know we cover everything. as to why the `.`, it's explained in the comment above ☝️

andresgutgon: Ok, just thinking that maybe is unexpected in a view something by mistake and be copied/deployed. Something that's not an asset

andresgutgon (approved):

packages/cli/src/lib/sync/syncStaticFiles/index.ts
csansoon: Yeah, we should make it clear in the docs that everything inside `views` is public.

PR #284, opened 2024-04-12
Use different event in embedding when the URL param change from inside the view

andresgutgon:
# What?
We were using the same event name when the param is changed from react host app than when is changed from inside view and this was causing an infinite loop 💀

### DIFF
--- a/.changeset/wild-eggs-drive.md
+++ b/.changeset/wild-eggs-drive.md
@@ -1,0 +1,6 @@
+---
+"@latitude-data/webcomponents": patch
+"@latitude-data/embedding": patch
+---
+
+Fix infinite loop when changing an input field. Use different event in embedding when url param is changed from inside view
--- a/packages/embedding/src/types.ts
+++ b/packages/embedding/src/types.ts
@@ -1,5 +1,6 @@
 export enum EmbeddingEvent {
   Run = 'latitude-run',
+  ChangeParamRequest = 'latitude-change-param-request',
   ParamsChanged = 'latitude-params-changed',
   Custom = 'latitude-custom',
 }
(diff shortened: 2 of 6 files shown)


.changeset/wild-eggs-drive.md
```diff
@@ -0,0 +1,6 @@
+---
+"@latitude-data/webcomponents": patch
+"@latitude-data/embedding": patch
+---
+
+Use different event in embedding when url param is changed from inside view
```
csansoon, line 6: We should say what bug this fixes, so someone who suffered it can see it has now been fixed.

packages/embedding/src/types.ts
```diff
@@ -21,6 +22,8 @@ export type EmbeddingEventData<T extends EmbeddingEvent> =
     ? { type: T } & RunEventData
     : T extends EmbeddingEvent.ParamsChanged
       ? { type: T; params: ParamsChangedEventData }
-      : T extends EmbeddingEvent.Custom
-        ? { type: T; data: unknown }
-        : never
+      : T extends EmbeddingEvent.ParamsChangedInView
```
csansoon, line 25: `ParamsChanged` and `ParamsChangedInView` are too similar.
Maybe use something like `ChangeParamsRequest` and `ParamsDidChange`

andresgutgon: That's is less clear imo. In both cases are params that were changed

csansoon (approved):

packages/embedding/src/types.ts
andresgutgon: Is more clear, I changed my mind 😂

Issue #255, opened 2024-04-08
Empty Input components polute the app url


geclos:
## How to reproduce
Enter an app with an empty Input component and the url automatically gets polluted with a `?param=$text:` query param. Example: https://file-over-app.fly.dev.

PR #293, opened 2024-04-15
Remove empty params when data apps load.

andresgutgon:
# What?
We want to avoid URLs in data apps with empty params like `?search=$text:`. Only when `search` has a non-empty value would appear in the URL

### Issue
https://github.com/latitude-dev/latitude/issues/255

### DIFF
--- a/.changeset/brown-mangos-marry.md
+++ b/.changeset/brown-mangos-marry.md
@@ -1,0 +1,5 @@
+---
+"@latitude-data/server": minor
+---
+
+Remove empty params from URL when data app loads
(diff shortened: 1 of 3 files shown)


geclos (approved):
@cesr remember this implies that you cannot set an input to empty string so it's kind of a limitation for a particular edge case the user might have

andresgutgon:
> [cesar](https://github.com/cesr) remember this implies that you cannot set an input to empty string so it's kind of a limitation for a particular edge case the user might have

No, the params are sent to the API even when is not visible in the URL. Is that correct @csansoon ?

PR #297, opened 2024-04-15
Improve latitude vite script tag detection and injection

csansoon:
Not all script tags in the view file can be used for injection. Only one main script tag is allowed in svelte, but script tags inside `<svelte:head>`, or with `type="module"` do not count as "main script tags". Implemented an actual XML parser to find the correct script tags for injection. 

Also, improved how we identify the methods that need to be transformed, as we now just take the raw expression (instead of trying to parse it and de-parse it to get the actual arguments), and just save the position for later precise replacement, instead of doing a find-and-replace over the whole document.

\+ added some tests for this script :)


apps/server/plugins/latitude/index.ts
csansoon, line 1: Also, this transformation is being done as a Vite plugin that gets ran on every request.
For performance, we could just perform this transformation when syncing the view files, from `index.html` to `+page.svelte`, right?

andresgutgon: Also needed when you start a fresh latitude server that doesn't have the files copied. For example if you clone the data app and start the server

csansoon: But that gets automatically synced anyways. The CLI still copies the `index.html` files into a structure of `+page.svelte` files. What I am saying is that that would be a good time to perform this transformation, so that it is static instead of being done every single request, since it will always return the exact same result for the same original file.

apps/server/plugins/latitude/index.test.ts
```diff
@@ -0,0 +1,141 @@
...
+      expect(scriptTag).toContain('console.log("Hello")') // Original code is still there
+      functions.forEach((fn) => expect(scriptTag).toContain(fn))
+
+      // Expressions have been replaced with the variable.
+      // This ensures the position is valid because it must respect the start and end '{}' characters.
+      const varRegex = /{\$__\w+_\d+}/g
+      expect(result.code).toMatch(varRegex)
+    })
+  })
+})
```
geclos, line 141: fantastic

apps/server/plugins/latitude/index.ts
geclos: why does it get run on every request? are u sure?

apps/server/plugins/latitude/index.ts
```diff
@@ -0,0 +1,185 @@
...
+    const startIndex = lastScript.get(0)!.startIndex
+    const position = html.indexOf('</script>', startIndex!)
+    return [ html, position ]
+  }
+
+  // If no script tag was found, create one at the end of the file
+  const scriptTag = '\n<script></script>'
+  const position = html.length + '<script>'.length + 1
+  return [ html + scriptTag, position ]
+}
```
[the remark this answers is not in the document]
geclos: stop with the ocd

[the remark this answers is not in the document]
andresgutgon: This is not ocd. Is a code style decision we can implement. I would say functions with export at the end.

apps/server/plugins/latitude/index.ts
csansoon: Yep. I was logging to console for debugging, and it only gets ran every single time I load the website.

geclos: but that is only in development, not a massive issue no?

andresgutgon (approved):

PR #298, opened 2024-04-16
Improve app build time by not loading all the connectors

andresgutgon:
# What?
We want to let users only install the connectors they need. This way deploy and installation is more lightwave

### TODO
- [x] Dynamically import connectors
- [x] Check when syncing sources (`*.yml`) if the connector `type` defined is included in the `root/package.json`. If not install it
- [x] Fix `runQuery` now `createConnector` is async ⚠️ ⚠️ 
- [x] How to keep in sync connectors installed in the data apps and the required versions used in `apps/server -> factory`
- [x] Think how to make this change to be backwards compatible or do a breaking change. Maybe not needed if the error message is clear enough. Will see.

## Do this needs to be a breaking change?
This question is here to be discussed. Imo no because we handle gracefully next time users update to this version of CLI it will install missing connectors when they run

1. `latitude dev`
2. `latitude build`

What do you think?

image


packages/connectors/factory/src/index.ts
```diff
@@ -50,42 +36,87 @@ export function createConnector(sourcePath: string): BaseConnector {
...
+  }
+}
+
+async function loadConnector(
+  packageName: string,
+) {
+  try {
+    const module = await import(packageName)
+    // Assuming the connector class is the default export of the module:
+    const ConnectorClass = module.default
```
andresgutgon, line 88: Making class connector default we don't need a name convention. This require change all connectors to have a default export.

packages/connectors/factory/src/index.ts
```diff
@@ -50,42 +36,91 @@ export function createConnector(sourcePath: string): BaseConnector {
...
+    // Now check if ConnectorClass is actually a constructor
+    if (
+      typeof ConnectorClass === 'function' &&
+      ConnectorClass.prototype.constructor
+    ) {
+      return ConnectorClass as any // cast to `any` temporarily; fix below
+    } else {
+      throw new Error(`Module ${packageName} does not export a constructor.`)
+    }
```
csansoon, line 102: Not only test that it is a constructor, but maybe test that it is for a subclass of `BaseConnector` too

csansoon: Also, this error message would be for the user, not for the package developer. So, instead of explaining that it "does not export a constructor", maybe just say "Module ... is not a valid connector for Latitude" or something like that. Users don't need to know what "exporting a constructor" even means.

packages/connectors/factory/src/index.ts
```diff
@@ -50,42 +36,91 @@ export function createConnector(sourcePath: string): BaseConnector {
...
-  if (!config?.type) throw new Error(`Missing 'type' in configuration`)
-  if (typeof config.type !== 'string')
-    throw new Error(`Invalid 'type' in configuration`)
+  return config
+}
+
+type PartialConfig = {
+  type?: string
+  npm_package?: { name: string }
+}
```
csansoon, line 45: We can actually reuse the `type`. If it's one of our reserved type names, use our package. If not, treat it like the name of the npm package.
Otherwise, there is no good expected behaviour if the user adds different values for both `type` and `npm_package`

andresgutgon: I did the `npm_package.name` because this way we can add `npm_package.version` and gives users the option to install diferent version of the package than latest. With this setup they even can install different versions of default adapters

packages/cli/src/lib/sync/shared/syncFiles.ts
```diff
@@ -35,11 +32,7 @@ export default function syncFiles({
       }
     }
 
-    if (strategy === 'symlink') {
```
andresgutgon, line 38: This strategy was not used by anyone.

packages/connectors/factory/src/index.ts
```diff
@@ -50,42 +38,82 @@ export function createConnector(sourcePath: string): BaseConnector {
...
-      return new SqliteConnector(path.dirname(sourcePath), details)
+      return '@latitude-data/sqlite-connector'
     case ConnectorType.Mssql:
-      return new MssqlConnector(path.dirname(sourcePath), details)
+      return '@latitude-data/mssql-connector'
     case ConnectorType.Databricks:
-      return new DatabricksConnector(path.dirname(sourcePath), details)
+      return '@latitude-data/databricks-connector'
+    default:
+      throw new Error(`Unsupported connector type: ${type}`)
```
andresgutgon, line 81: For now we raise an error. When we want to give publicity to custom connectors we just return the `type`

packages/connectors/factory/src/index.ts
```diff
@@ -50,42 +38,82 @@ export function createConnector(sourcePath: string): BaseConnector {
...
+  }
+}
+
+async function loadConnector(packageName: string) {
+  try {
+    const module = await import(/* @vite-ignore */ packageName)
```
andresgutgon, line 87: Vite complained in runtime and suggested this comment

packages/connectors/factory/src/missingConnector.test.ts
```diff
@@ -0,0 +1,39 @@
+import * as fs from 'fs'
+
+import { it, describe, beforeEach, expect } from 'vitest'
+import { vi } from 'vitest'
+import { createConnector } from './index'
+
+vi.mock('@latitude-data/postgresql-connector', () => ({
+  default: 'a simple string',
+}))
```
andresgutgon, line 9: Mocking packages is not straight forward. If you mock as a string the rest of the tests fail. For that I made an independent test to test this case

packages/cli/src/lib/sync/syncQueries/index.ts
```diff
@@ -12,77 +12,46 @@ import {
...
-      .relative(rootDir, srcPath)
-      .replace(/^queries\/?/, '')
-
-    let destPath
-
-    if (srcPath.endsWith('.csv')) {
-      destPath = path.join(destinationCsvsDir, relativePath)
-    } else if (
-      srcPath.endsWith('.sql') ||
-      srcPath.endsWith('.yml' || srcPath.endsWith('.yaml'))
```
andresgutgon, line 77: Here was a bug. Can you see it?

packages/connectors/factory/package.json
```diff
@@ -25,19 +25,21 @@
     "typescript": "^5.2.2",
     "vitest": "^1.2.2"
   },
-  "dependencies": {
+  "peerDependencies": {
```
andresgutgon, line 28: 🌳 THIS IS THE MAIN CHANGE

From now on all connectors are not installed. Only setup the relation between the factory and those connectors so we can read the package.json from the factory and install the right version on the data app whenever the user updates `apps/server`. This should keep in-sync connectors and `apps/server` as before.

packages/cli/src/lib/sync/syncQueries/ensureConnectorInstalled.ts
```diff
@@ -0,0 +1,202 @@
...
+} from '@latitude-data/connector-factory'
+import { APP_FOLDER } from '$src/commands/constants'
+
+function installErrorMessage({ npmPackage }: { npmPackage: string }) {
+  return `
+🚨 Failed to install connector ${npmPackage}
+
+Please run 'npm init -y' in the app directory.
+And then run 'npm install ${npmPackage}' to install the connector.
+`
```
csansoon, line 17: Maybe just run `npm init -y && npm install ${npmPackage}`

csansoon: Wait what if npm was already setup and the problem is just that the module is not installed? Can the `npm init -y` fail or warn anything in that case?

csansoon (approved):

packages/connectors/mssql/src/index.ts
```diff
@@ -22,7 +22,7 @@ export type ConnectionParams = {
   }
 }
 
-export class MssqlConnector extends BaseConnector {
+export default class MssqlConnector extends BaseConnector {
```
geclos, line 25: hmm this is a breaking change though... should we avoid it? to no include more entropy than needed to this commit

andresgutgon: Agree but dynamic importing is way easier this way. Read the comment around it where I use it. Also these components are only used by us

geclos: > Also these components are only used by us

We... don't really know this. Let's assume it ain't the case as a baseline.

packages/cli/src/lib/sync/syncQueries/index.ts
geclos: what's the bug?

packages/connectors/factory/src/index.ts
geclos: hmm ya, makes sense, ok let's do it but do a major update to all connectors please 🙏🏼

packages/cli/src/lib/sync/syncQueries/ensureConnectorInstalled.ts
```diff
@@ -3,37 +3,40 @@ import spawn from '$src/lib/spawn'
...
+import config from '$src/config'
+import isSourceFile from '$src/lib/isSourceFile'
 
 function installErrorMessage({ npmPackage }: { npmPackage: string }) {
   return `
 🚨 Failed to install connector ${npmPackage}
 
 To install manually the connector run:
-'npm init -y && npm install ${npmPackage}' in the app directory.
+'npm install --save ${npmPackage}' in the app directory.
```
geclos, line 20: save flag will automatically add the package.json if it does not exist, no point in the npm init since we don't care about any other prop other than dependencies.

packages/cli/src/lib/sync/syncQueries/ensureConnectorInstalled.ts
```diff
@@ -3,37 +3,40 @@ import spawn from '$src/lib/spawn'
...
   return new Promise((resolve) => {
     spawn(
       'npm',
-      ['install', `${npmPackage}@${npmVersion}`],
+      ['install', '--save', `${npmPackage}@${npmVersion}`],
       {
-        cwd: rootDir,
-        stdio: 'inherit',
+        cwd: config.rootDir,
+        stdio: 'pipe',
```
geclos, line 39: I've opted for logging a bit less here, I think it's a better experience during development, but open to changing my opinion honestly 🤷🏼

packages/cli/src/lib/sync/syncQueries/ensureConnectorInstalled.ts
```diff
@@ -3,37 +3,40 @@ import spawn from '$src/lib/spawn'
...
-  rootDir,
   npmPackage,
   npmVersion,
 }: {
-  rootDir: string
   npmPackage: string
   npmVersion: string
 }) {
   const errorMessage = installErrorMessage({ npmPackage })
+  console.log(colors.yellow(`Installing connector ${npmPackage}...`))
```
geclos, line 32: important to warn people why an install has suddenly started!

packages/cli/src/lib/sync/syncQueries/ensureConnectorInstalled.ts
```diff
@@ -66,77 +69,24 @@ function installConnector({
...
-  await maybeInitPackageJson({ rootDir, npmPackage })
-
-  const file = fs.readFileSync(`${rootDir}/package.json`, 'utf8')
-  const packageJson = JSON.parse(file)
-  const dependencies = packageJson.dependencies || {}
-  return dependencies[npmPackage]
+  try {
+    const config = loadConfig({ sourcePath: srcPath })
+    return getConnectorPackage(config.type as ConnectorType)
+  } catch (e) {
```
geclos, line 76: This was failing if i added an empty source. Also removed the boolean prop as it was leaking implementation details.

packages/cli/src/lib/sync/syncQueries/ensureConnectorInstalled.ts
```diff
@@ -66,77 +69,24 @@ function installConnector({
...
 }
 
 function getConnectorVersionInLatitudeApp({
-  rootDir,
   npmPackage,
 }: {
-  rootDir: string
   npmPackage: string
 }) {
-  const factoryPackagePath = `${rootDir}/${APP_FOLDER}/node_modules/@latitude-data/connector-factory`
```
geclos: never construct the path to the folder, it's in config

packages/cli/src/lib/sync/syncQueries/ensureConnectorInstalled.ts
```diff
@@ -66,77 +69,24 @@ function installConnector({
...
-            onError({ message: errorMessage })
-            process.exit(1)
-          } else {
-            resolve(true)
-          }
-        },
-      },
-    )
-  })
-}
```
geclos: no needed

packages/cli/src/lib/sync/syncQueries/ensureConnectorInstalled.ts
```diff
@@ -66,77 +69,24 @@ function installConnector({
...
-}: {
-  npmPackage: string
-  rootDir: string
-}) {
-  await maybeInitPackageJson({ rootDir, npmPackage })
-
-  const file = fs.readFileSync(`${rootDir}/package.json`, 'utf8')
-  const packageJson = JSON.parse(file)
-  const dependencies = packageJson.dependencies || {}
-  return dependencies[npmPackage]
```
geclos: opted to always run the install instead of doing the early return. The install is a noop if all modules are already installed and it's less code. Also this is a bug since i can have the package.json but not have the dependency installed.

packages/cli/src/lib/sync/syncQueries/ensureConnectorInstalled.ts
```diff
@@ -146,57 +96,25 @@ function getConnectorVersionInLatitudeApp({
...
+  type,
 }: {
-  rootDir: string
   srcPath: string
-  isSource: boolean
+  type: 'add' | 'change' | 'unlink'
 }) {
-  if (!isSource) return true
+  if (!isSourceFile(srcPath)) return
+  if (type !== 'add' && type !== 'change') return
```
geclos, line 110: the cli was breaking when deleting a source file

packages/cli/src/lib/sync/syncQueries/index.ts
```diff
@@ -13,26 +13,19 @@ import {
...
-  return srcPath.endsWith('.yml') || srcPath.endsWith('.yaml')
-}
+import isSourceFile from '$src/lib/isSourceFile'
 
 function buildDestPath({
-  rootDir,
   srcPath,
   destinationQueriesDir,
   destinationCsvsDir,
-  isSource,
```
geclos: leaky boolean prop

packages/cli/src/lib/sync/syncQueries/ensureConnectorInstalled.ts
andresgutgon: I guess you mocked the tests

packages/cli/src/lib/sync/syncQueries/ensureConnectorInstalled.ts
andresgutgon: Better check if `unlink` no?

packages/cli/src/lib/sync/syncQueries/syncQueriesAndCsvs.test.ts
```diff
@@ -24,6 +24,8 @@ vi.mock('../shared/syncFiles', () => ({
   default: vi.fn(),
 }))
 
+vi.mock('$src/config', () => ({ default: { rootDir: '/fake/rootDir' } }))
```
andresgutgon, line 27: Yep ok

andresgutgon (commented):
👍 Awesome thanks!

packages/cli/src/lib/sync/syncQueries/ensureConnectorInstalled.ts
geclos: I'm afraid there might be other events I don't know about that's why I prefer to check for the events I want to react to

apps/server/package.json
```diff
@@ -63,6 +63,8 @@
     "@latitude-data/query_service": "workspace:*",
     "@latitude-data/svelte": "workspace:*",
     "cheerio": "1.0.0-rc.12",
-    "lodash-es": "^4.17.21"
+    "lodash-es": "^4.17.21",
+    "@latitude-data/clickhouse-connector": "^0.1.0",
+    "@latitude-data/duckdb-connector": "^1.0.3"
```
andresgutgon, line 68: 👀 We have this problem when we install at development while developing connectors ends here.

geclos: yepa

PR #300, opened 2024-04-17
feature: Latitude PaaS integration

geclos:
# WHAT

This commit implements the first commands to interact with the Latitude PaaS. Specifically:
- `latitude signup`: To signup to latitude from the CLI
- `latitude login`: To login to your existing account
- `latitude logout`: To logout from your existing account
- `latitude deploy`: To deploy your application to the Latitude PaaS
- `latitude destroy`: To destroy your application from the Latitude PaaS
- `latitude cancel`: To cancel an ongoing deployment to the Latitude PaaS


packages/cli/src/commands/login/index.ts
```diff
@@ -0,0 +1,45 @@
...
+      return
+    } else {
+      const { response, responseBody } = res
+      if (response.statusCode === 200) {
+        try {
+          const response = JSON.parse(responseBody)
+          const jwt = response.token
+          if (jwt) {
+            configStore.set('jwt', jwt)
+            console.log(chalk.green('Login successful! JWT stored.'))
```
andresgutgon, line 28: `Login successfully as ${email} 🎉`

packages/cli/src/commands/login/index.ts
```diff
@@ -0,0 +1,45 @@
...
+          } else {
+            throw new Error('JWT not found in response')
```
andresgutgon, line 30: No need to talk about JWT. Say something more generic

packages/cli/src/commands/signup/index.ts
```diff
@@ -0,0 +1,44 @@
...
+      const { response, responseBody } = res
+      if (response.statusCode === 200) {
+        try {
+          const response = JSON.parse(responseBody)
+          const jwt = response.token
+          if (jwt) {
+            configStore.set('jwt', jwt)
+            console.log(chalk.green('Signup successful! JWT stored.'))
+          } else {
+            throw new Error('JWT not found in response')
```
andresgutgon, line 29: ditto

andresgutgon (approved):

packages/cli/src/commands/signup/index.ts
```diff
@@ -0,0 +1,44 @@
...
+      console.error('Signup request failed:', err.message)
+    } else {
+      const { response, responseBody } = res
+      if (response.statusCode === 200) {
+        try {
+          const response = JSON.parse(responseBody)
+          const jwt = response.token
+          if (jwt) {
+            configStore.set('jwt', jwt)
+            console.log(chalk.green('Signup successful! JWT stored.'))
```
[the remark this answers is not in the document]
andresgutgon: This deserves a boxed message where you can explain that they can use deploy command.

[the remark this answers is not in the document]
geclos: I don't agree I think we overuse boxed messages. If you sign up I think we can assume people know why they are signing up.

andresgutgon (approved):

packages/cli/src/cli.ts
```diff
@@ -87,6 +94,37 @@ CLI.command('credentials')
...
+
+CLI.command('destroy')
+  .describe('Destroy a deployed Latitude app')
+  .action(destroyCommand)
+
+CLI.command('cancel')
+  .describe('Cancels an ongoing deployment')
+  .action(cancelCommand)
+
+CLI.command('poke').describe('Poke Latitude server').action(pokeCommand)
```
andresgutgon, line 126: I would organize all these commands under `commands/cloud` folder. It's easier to understand that all belong to the same area

packages/cli/src/commands/cancel/index.ts
```diff
@@ -0,0 +1,44 @@
...
+          )
+
+          process.exit(1)
+        }
+      }
+    },
+  )
+}
+
+export default tracked('cancelCommand', loggedIn(cancel))
```
andresgutgon, line 44: Do we really need to check this on the frontend. The backend is going to fail anyway

geclos: this is actually checking in the backend, it's just to ensure we display a nice error in the frontend if the user is not logged in.

andresgutgon: Yes, but I mean the deploy endpoint could return this nice error. No need for 2 roundtrips.

geclos: hmm yup good point i was tunnel visioned

PR #307, opened 2024-04-18
Fix: server was opening in wrong port

csansoon:


andresgutgon (dismissed):

packages/cli/src/commands/dev/runDev.ts
```diff
@@ -44,6 +44,7 @@ export async function runDevServer(
   const args = [
     'run',
     'dev',
+    '--',
     '--strictPort',
     `--port=${appPort}`,
     `--host=${host}`,
```
csansoon, line 50: Arguments when opening the server were being ignored.
This means that the `--open` flag was being ignored too. Now the browser is open by default when starting the server, although that was not happening before. Do we keep it that way?

andresgutgon: I think we decided not to open by default

csansoon:
```
  Description
    Launch the local Latitude development environment

  Usage
    $ latitude dev [options]

  Options
    --open            Open the Latitude app in your browser, Default: yes. Options: yes, no
    ...
```
It is enabled by default

csansoon: Okay, changed it to not be enabled by default :)

geclos: no no! it's ok to be opened by default! but just in production, not in development

packages/cli/src/cli.ts
```diff
@@ -53,10 +53,7 @@ CLI.command('build')
 
 CLI.command('dev')
   .describe('Launch the local Latitude development environment')
-  .option(
-    '--open',
-    'Open the Latitude app in your browser, Default: yes. Options: yes, no',
```
geclos, line 58: default opened 🙏🏼 like nextjs and company

packages/cli/src/commands/dev/runDev.ts
```diff
@@ -44,6 +44,7 @@ export async function runDevServer(
   const args = [
     'run',
     'dev',
+    '--',
```
geclos, line 47: good catch 👌🏼 

any other npm run commands that might be missing this in the cli?

packages/cli/src/lib/spawn/index.ts
```diff
@@ -13,6 +13,17 @@ export default function betterSpawn(
...
+    // Fix npm ignoring command flags
+    const firstFlagIndex = args.findIndex((arg) => arg.startsWith('--'))
+    if (firstFlagIndex !== -1) {
+      args = args
+        .slice(0, firstFlagIndex)
+        .concat(['--'])
+        .concat(args.slice(firstFlagIndex))
+    }
+  }
+
```
csansoon, line 26: @geclos What about this? Feel like its overcomplicating things, but this would fix any future bug like this.

geclos: hmm I'd say better to leave it to the consumer I don't think it's the responsibility of this lib to defend against idiosyncrasies of npm

PR #319, opened 2024-04-23
New template name

cesr:
From netflix to duckdb

### DIFF
--- a/packages/cli/src/commands/start/cloneTemplate.ts
+++ b/packages/cli/src/commands/start/cloneTemplate.ts
@@ -6,7 +6,7 @@ import { TemplateUrl } from './questions'
 
 const TEMPLATE_URL: Record<TemplateUrl, string> = {
   [TemplateUrl.default]: `${LATITUDE_GITHUB_SLUG}/template`,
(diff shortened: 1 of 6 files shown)


packages/cli/src/commands/start/cloneTemplate.ts
```diff
@@ -6,7 +6,7 @@ import { TemplateUrl } from './questions'
 
 const TEMPLATE_URL: Record<TemplateUrl, string> = {
   [TemplateUrl.default]: `${LATITUDE_GITHUB_SLUG}/template`,
-  [TemplateUrl.netflix]: `${LATITUDE_GITHUB_SLUG}/sample-netflix`,
+  [TemplateUrl.duckdb]: `${LATITUDE_GITHUB_SLUG}/sample-duckdb`,
```
andresgutgon, line 9: Why we have these submodules?

andresgutgon: No no, my mistake. forget it

andresgutgon (approved):

PR #324, opened 2024-04-24
Improved docs for SQL syntax

csansoon:
I removed the docs we had on basic regular SQL syntax (I think that we can skip that part, since our userbase most likely already knows basic SQL syntax, and we already link to each source's documentation). 
Instead, in that same place I added documentation on our custom syntax: what it is, how it works, and what can do.

### DIFF
--- a/docs/queries/advanced/interpolation.mdx
+++ b/docs/queries/advanced/interpolation.mdx
@@ -5,56 +5,115 @@ description: 'Insert values into SQL queries without parametization'
 
 ## Introduction
 
-Interpolation refers to inserting values directly into the SQL code but without parametrization.
+Interpolation refers to inserting values directly into the SQL code without parametrization.
 
-By default, all of Latitude's logic is parametrized to independently send the query and values to the database and to detect possible SQL injection. Interpolating a value breaks the parametrization.
+In order to prevent SQL injection attacks, Latitude [parameterizes](https://cheatsheetseries.owasp.org/cheatsheets/Query_Parameterization_Cheat_Sheet.html)
+all queries to your db, by default. Parameter interpolation bypasses this protection and users should use it at their own risk. Read more about it
+in the [SQL Syntax](/queries/basics/sql-syntax-basics#adding-values-to-the-query) section.
 
 <Warning>Direct interpolation can lead to SQL injection vulnerabilities, allowing users to perform critical actions such as deleting a table.</Warning>
+
+## Why use interpolation?
+
+There are some cases where you might want to interpolate a value directly into the SQL query.
+- **unparameterizable values**: Depending on the database engine, some values cannot be parameterized. For example, table names cannot be parameterized in PostgreSQL.
+
+  For example:
+  ```sql
+    SELECT
+      user_name as {param('column_name')}
+    FROM users
+  ```
+  Running this query in PostgreSQL will throw an error because the column name cannot be parameterized. To fix it, we must interpolate the column name directly into the query:
+  ```sql
+    SELECT
+      user_name as {interpolate(param('column_name'))}
+    FROM users
+  ```
+
+- **Dynamic query keywords**: Most database engines only let you parameterize values, not keywords. For example, you cannot parameterize the `DESC` or `ASC` keyword in some engines.
+  ```sql
+    SELECT *
+    FROM users
+    ORDER BY users.id {param('order')}
+  ```
+  In order to fix this, you can interpolate the keyword directly into the query:
+  ```sql
+    SELECT *
+    FROM users
+    ORDER BY users.id {interpolate(param('order'))}
+  ```
 
 ## Syntax
 
 `interpolate()` allows you to send any parenthesized logic directly to the database. Let's look at some examples:
 
-```
+```sql
 interpolate({your_logic_here})
 ```
 
-- **Variable without interpolation**
+Take into account that interpolating any value will just add it straight to the query. This means that interpolating a string will not add quotes around it, so you must add them manually if needed.
 
-    ``` sql variable.sql
-    {companyFiltered = 'Latitude'}
+```sql
+SELECT *
+FROM table
+WHERE company = \'{interpolate('company_name')}\'
+```
 
-    SELECT *
-    FROM table
-    WHERE company = {companyFiltered}
-    ```
-    Will send the query parametized:
-    ``` sql
-    SELECT *
-    FROM table
-    WHERE company = $1
-    ```
-    And the value of `$1` independently:
-    ``` js
-    $1 = 'Latitude'
-    ```
-    This allows databases to check the values and see if there is a security vulnerability.
+Here we are also adding quotes around the interpolation to make sure the query is correctly formatted. We are also escaping the quotes, because otherwise everything inside a string is just considered a regular string and not compiled by Latitude.
 
-- **Variable with interpolation**
+## Example
 
-    ``` sql variable.sql
-    {companyFiltered = 'Latitude'}
+### Without interpolation
 
-    SELECT *
-    FROM table
-    WHERE company = {interpolate({companyFiltered})}
-    ```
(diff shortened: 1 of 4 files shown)


andresgutgon (approved):

docs/queries/advanced/interpolation.mdx
```diff
@@ -5,56 +5,102 @@ description: 'Insert values into SQL queries without parametization'
 
 ## Introduction
 
-Interpolation refers to inserting values directly into the SQL code but without parametrization.
+Interpolation refers to inserting values directly into the SQL code without parametrization.
 
-By default, all of Latitude's logic is parametrized to independently send the query and values to the database and to detect possible SQL injection. Interpolating a value breaks the parametrization.
+By default, all of Latitude's logic is safely parametrized into the query so that query and values are sent independently to your database. This is done for safely reasons,
```
geclos, line 10: In order to prevent SQL injection attacks, Latitude [parametrizes](https://cheatsheetseries.owasp.org/cheatsheets/Query_Parameterization_Cheat_Sheet.html) all queries to your db, by default. Parameter interpolation bypasses this protection and users should use it at their own risk.

geclos (dismissed):

docs/queries/advanced/interpolation.mdx
```diff
@@ -5,56 +5,102 @@ description: 'Insert values into SQL queries without parametization'
...
+to not avoid any SQL Injection or similar vulnerabilities. Interpolating a value skips this parametrization process. Read more about it in [Adding values to the query](/queries/basics/sql-syntax-basics#adding-values-to-the-query).
 
 <Warning>Direct interpolation can lead to SQL injection vulnerabilities, allowing users to perform critical actions such as deleting a table.</Warning>
 
+## Why use interpolation?
+
+There are some cases where you might want to interpolate a value directly into the SQL query.
+- **Unparametrizable values**: Depending on the database engeing, some values cannot be parametrized. For example, table names cannot be parametrized in PostgreSQL.
```
geclos, line 18: engine*

docs/queries/advanced/interpolation.mdx
```diff
@@ -5,56 +5,102 @@ description: 'Insert values into SQL queries without parametization'
...
-    SELECT *
-    FROM table
-    WHERE company = $1
-    ```
-    And the value of `$1` independently:
-    ``` js
-    $1 = 'Latitude'
-    ```
-    This allows databases to check the values and see if there is a security vulnerability.
+Here we're also adding quotes around the interpolation to make sure the query is correctly formatted. We are also escaping the quotes, because otherwise everything inside a string is just considered a regular string and not compiled by Latitude.
```
geclos, line 62: Here, we are* adding

docs/queries/advanced/interpolation.mdx
```diff
@@ -5,56 +5,102 @@ description: 'Insert values into SQL queries without parametization'
...
+  WHERE company = {companyName}
+  ```
+
+**Compiled query:**
+  ```sql
+  SELECT *
+  FROM table
+  WHERE company = $1
+  ```
+  `$1 = 'Latitude'`
```
geclos, line 83: comment here that this parameter is sent to the db separately, which prevents any possible SQL injection attack.

docs/queries/advanced/interpolation.mdx
cesr: parameterized*

csansoon: I'll add this to the `SQL Syntax` documentation instead where it is explained in more detail, and then link it here as a reference.

docs/queries/basics/sql-syntax-basics.mdx
```diff
@@ -27,65 +25,109 @@ You can check the documentation for each database or repository.
...
-  ```
-  FROM table_name
-  ```
+With Latitude, you can use additional logic to make your queries more dynamic and powerful.
 
-- **Columns**: If you want to refer to a specific column, you can use the table name followed by the column with a period in the middle. 
-  ```
-  JOIN second_table where second_table.column_name
-  ```
+In any query, you can use **Logic Blocks**, which are always enclosed in curly brackets `{}`. These blocks can contain any logic you need, and they can be used
```
geclos, line 32: no any logic no? can't declare functions, etc... it's not regular javascript inside

docs/queries/basics/sql-syntax-basics.mdx
```diff
@@ -27,65 +25,109 @@ You can check the documentation for each database or repository.
...
+to modify the query, add conditions, or even create loops.
```
geclos, line 33: weird jump line from previous line

docs/queries/basics/sql-syntax-basics.mdx
```diff
@@ -27,65 +25,109 @@ You can check the documentation for each database or repository.
...
 
-## Reuse Queries
+Before sending the query to the database, Latitude compiles the logic in your query, and then sends the compiled query to the database. Your database will never
```
geclos, line 35: weird jumpline

docs/queries/advanced/interpolation.mdx
csansoon: Oh nono, I'm just adding the link to the SQL Syntax docs and adding that paragraph to Interpolation too.

docs/queries/basics/sql-syntax-basics.mdx
```diff
@@ -27,65 +25,109 @@ You can check the documentation for each database or repository.
...
+AND created_at > {param('register_date')} -- function
 ```
-Instead of sending the code as it is, we send the following to the database
 
-``` sql parametized_query
-SELECT
-  $1 AS name
-  $2 AS name_and_surname
+By default, all values will always be safely parameterized. This means that Latitude will not directly include the value to the query, but instead
+it adds a reference to the value and sends the value separately to the database.
```
geclos, line 87: Link to where we show parametrized queries in the docs

geclos: ah wait so this is the section...

geclos: ok it's fine then

geclos (approved):
i see weird jumplines in your md make sure it does not affect the rendered view

docs/queries/basics/sql-syntax-basics.mdx
```diff
@@ -27,65 +25,110 @@ You can check the documentation for each database or repository.
...
+
+`$1: 'Latitude'`
+`$2: 18`
+`$3: '2021-01-01'`
+
+How the actual reference looks like depends on your source type, all of them are compatible with this feature. This is a security measure to prevent
+SQL injection and other vulnerabilities.
+
+If you need to interpolate a value directly into the query bypassing this parameterization, you can use the `interpolate` function. Read more about
+it in the [Interpolatation](/queries/advanced/interpolation) section.
```
cesr, line 104: interpopolatatatitirititation

docs/queries/basics/sql-syntax-basics.mdx
csansoon: Any Latitude logic, which is explained below how it looks and what it can do.

I am not specifying that this is javascript or any javascript-related logic (although it is). Without knowing it, it just looks like some generic code syntax.

docs/queries/basics/sql-syntax-basics.mdx
csansoon: Oh that's just so that it looks readable from the code editor. Mintlify actually ignores single line breaks for this reason.
To actually add a line break to the docs you must use at least two in a row.

docs/queries/basics/sql-syntax-basics.mdx
csansoon: ditto

PR #331, opened 2024-04-25
Support for react components

csansoon:
Added support for built-in out-of-the-box components in the React package.

Since the React package is now more complex and allows for multiple use cases, a new tab in the documentation has been added. Currently, the "React" documentation will have the following explanations:
- Installation (not only it needs to be added to a React project, but it needs to be wrapped in a Provider)
- Embedding views
- Running queries using `useQuery`
- Using built-in components

For the sake of the component support, some simple components for different use cases have been added. More will be added in a different PR.


examples/sample-react/src/examples/Components/index.tsx
```diff
@@ -0,0 +1,33 @@
+import { Card, Text } from "@latitude-data/react"
```
andresgutgon, line 1: Instead of doing this setup storybook to have more organized visualization of the UI elements

csansoon: yeah I'm using this one just to temporarily test them while I'm developing. But yeah, I should definitely add a storybook for each component...

docs/mint.json
```diff
@@ -209,6 +213,21 @@
...
+        "react/getting-started/installation",
+        "react/getting-started/embedding-views"
+      ]
+    },
+    {
+      "group": "Running queries",
+      "pages": [
+        "react/running-queries/use-query",
+        "react/running-queries/query-result"
+      ]
```
andresgutgon, line 230: Aren't docs in another PR?

packages/client/react/components.json
```diff
@@ -0,0 +1,17 @@
+{
+  "$schema": "https://ui.shadcn.com/schema.json",
+  "style": "new-york",
```
andresgutgon, line 3: Make sure is the same style that we use in svelte/components.json

.changeset/lemon-kids-tickle.md
```diff
@@ -0,0 +1,7 @@
+---
+"@latitude-data/svelte": minor
+"@latitude-data/react": minor
+"@latitude-data/client": minor
+---
+
+Added support for built-in components for Latitude's React package.
```
andresgutgon, line 7: Is this going to go to next changeset? I guess yes because is pointing to `next` branch.

andresgutgon (approved):

andresgutgon:
Make sure when deployed next- version of this package that embedding keeps working. Use new version of this package. 

You can fork this repo and update the version
https://github.com/andresgutgon/latitude-embed/blob/main/package.json#L14

.changeset/lemon-kids-tickle.md
csansoon: Yes :)

docs/mint.json
csansoon: Documentation about components will be done in a different PR. This are already existing features, which I am porting and adding to the new "React" tab here.

PR #332, opened 2024-04-25
Unify api parameter when deploying and call it always app

andresgutgon:
# What?
What is in the title

This part is dependent on the server change but I think is fine if we merge it no?

### DIFF
--- a/packages/cli/src/commands/cloud/deploy/index.ts
+++ b/packages/cli/src/commands/cloud/deploy/index.ts
@@ -7,12 +7,12 @@ import pushDockerImage from './docker/push'
 import tracked from '$src/lib/decorators/tracked'
 import { request, sseRequest } from '$src/lib/server'
 
-async function deploy({ name, digest }: { name: string; digest: string }) {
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/cloud/deploy/index.ts
```diff
@@ -74,14 +74,14 @@ async function deployCommand(
...
-  const name = latitudeJson.data.name!
+  const app = latitudeJson.data.name!
 
-  console.log(`Deploying ${name}...`)
+  console.log(`Deploying ${app}...`)
 
   const { url, latestImage } = await request({
     method: 'POST',
     path: '/api/ecr/repositories/find-or-create',
-    data: JSON.stringify({ name }),
```
geclos, line 84: this needs a change in the server api also

andresgutgon: Yes, that's what I said in the description. This line is the one that triggered this change

geclos: let's wait for the server change before merging this

PR #341, opened 2024-04-25
feature: update root dependencies on latitude update

geclos:
Well, this is embarrasing... There's a dependency between latitude server and connectors (obviously), and there's really no mechanism to ensure compatibility between both. So, a user could install version X of latitude server and install version Y of a connector, and they wouldn't work together because there's no mechanism to ensure compatibilty at install time.

For now, this commit implements an npm update of root packages on `latitude update`. We are optimistically assuming most users will update to the latest version so it's fair to assume an npm update will be the wanted action most of the times. Far from a perfect solution.


packages/cli/src/cli.ts
```diff
@@ -43,6 +43,10 @@ CLI.command('start')
 
 CLI.command('update')
   .describe('Update latitude app. You can define the version with --version')
+  .option(
+    '--force',
+    'Forces update of dockerignore and Dockerfile files to the latest version',
+  )
   .option('--fix', 'Installs the version defined in latitude.json')
   .action(updateCommand)
```
csansoon, line 51: Why does the user need to add a force flag to update the Dockerfile?

csansoon (approved):

PR #343, opened 2024-04-26
Added custom tailwind prefix

csansoon:
'@latitude-data/svelte' package can now be used in svelte projects without needing to configure tailwind.

- All classes in `client/core` are defined with a `lat-` prefix to avoid conflict with existing tailwind config in projects.
- `client/svelte` compiles all its used classes from core using tailwind (with the prefix) and exports a .css file
- `app/server` imports the .css file to all views. Now the app does not NEED to have tailwind.
- Additionally, `app/server` also has tailwind configured so that regular classes without the prefix still work too.


packages/client/svelte/src/lib/ui/theme-wrapper/switcher/index.ts
```diff
@@ -1,7 +1,2 @@
 import Root from './index.svelte'
-
-export function updateInlineStyles() {
-  import('@latitude-data/client/latitude.css')
-}
-
```
andresgutgon, line 6: lol I don't even remember what was doing this

geclos (approved):
awesom

PR #344, opened 2024-04-26
patch: add package version to dockerfile

geclos:
This commmit adds the cli package version to the dockerfile, this way we ensure that whenever the cli version changes and the user updates the project the dockerfile downloads the correct cli version for building the production build.


andresgutgon (approved):
Nice

PR #350, opened 2024-04-29
feature: added e2e tests for latitude run

geclos:
E2E test to ensure the query running actually works.


packages/cli/tests/e2e/run.cjs
```diff
@@ -0,0 +1,11 @@
+const { spawnSync } = require('child_process')
+const { writeFileSync } = require('fs')
+
+export default function run () {
+  writeFileSync('queries/nested/source.yml', 'type: duckdb')
```
andresgutgon, line 5: Create in `/tmp` maybe. This way these files are cleanup over time

PR #359, opened 2024-04-30
Improved theming api and docs

csansoon:
## Describe your changes

Following the components integration in the React package, we need a better API for custom theming, along with a good documentation.


packages/client/core/src/theme/skins/buildCssVariables.ts
```diff
@@ -1,26 +1,37 @@
...
+  const result = {} as T;
+  for (const key in defaultObject) {
+    if (typeof defaultObject[key] === 'object') {
+      result[key] = mergeObjects(defaultObject[key], partialObject[key] || {});
+    } else {
+      result[key] = partialObject[key] || defaultObject[key];
+    }
+  }
+  return result;
+}
```
csansoon, line 37: Merging two `a` and `b` objects like `c = { ...a, ...b }` may lose information when dealing with nested objects.

For example, if we have these two objects:
```js
defaultObject = {
  a: 1,
  b: 2,
  c: {
    c1: 1,
    c2: 2
  }
}

partialObject = {
  c: {
    c2: 4
  }
}
```

Merging them would result in an object where the "c.c1" property has been removed, since it completely replaces `defaultObject.c` with `partialObject.c`, independently of its content.

This method recursively merges the two objects, fixing this problem. This is used on themes because they both have two different `light` and `dark` nested objects, and we will add a new option for styling echarts in each of them.

geclos (approved):

docs/react/components/theming.mdx
```diff
@@ -0,0 +1,374 @@
...
+
+  function App() {
+    return (
+      <LatitudeProvider theme={myTheme}>
+        <AppContent />
+      </LatitudeProvider>
+    );
+  }
+  ```
+  ```tsx TypeScript
```
andresgutgon, line 108: Not sure if add much value the example in both js and ts. This is very simple example

packages/client/core/src/theme/skins/types.ts
```diff
@@ -18,7 +18,8 @@ export type ThemeColor = {
   border: string
   input: string
   ring: string
-  radius?: string
+  radius: string
+  echarts: object
```
andresgutgon, line 22: Echarts doesn't have types for this? I think they don't but can you check?

packages/client/svelte/src/lib/internal/ui/chart/_Echarts.svelte
```diff
@@ -64,27 +66,32 @@
...
   }
 
   // TODO: Implement theming
-  const theme = client.ui.chart.THEMES.latitude
   export let options: Props['options']
   export let width: Props['width'] = undefined
   export let height: Props['height'] = undefined
   export let locale: Props['locale'] = undefined
 
+  const currentTheme = derived(themeConfig, ($themeConfig) => $themeConfig.light.echarts!)
```
andresgutgon, line 78: Did you find how to detect dark mode?

andresgutgon (approved):

packages/client/core/src/theme/skins/buildCssVariables.ts
```diff
@@ -1,26 +1,23 @@
...
-  // Iterate over dark theme properties
-  for (const [key, value] of Object.entries(dark)) {
-    cssVariables += `  --${LATITUDE_PREFIX}-${key}: ${value};\n`
-  }
+const LATITUDE_PREFIX = 'lat-'
+
+function mapThemeToCssVariables(obj: Record<string, unknown>): string {
+  return Object.entries(obj)
+    .map(([key, value]) => {
+      if (typeof value !== 'string' && typeof value !== 'number') return
```
geclos, line 8: you are gonna join null values here

geclos (approved):

packages/client/core/src/theme/skins/buildCssVariables.ts
csansoon: good catch!

packages/client/core/src/theme/skins/types.ts
csansoon: They don't :(
https://github.com/apache/echarts/issues/18770

packages/client/core/src/theme/skins/buildCssVariables.ts
geclos: add tests

PR #367, opened 2024-05-03
⭐️ Little GitHub stars banner in start/deploy commands in CLI

andresgutgon:
## Describe your changes
Little GitHub stars banner in start/deploy commands in CLI ⭐️

image

### DIFF
--- a/.changeset/quiet-dragons-fix.md
+++ b/.changeset/quiet-dragons-fix.md
@@ -1,0 +1,5 @@
+---
+"@latitude-data/cli": minor
+---
+
+Little GitHub stars banner in deploy and start commands in CLI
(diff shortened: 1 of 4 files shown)


geclos:
are we convinced we want to show this on start command? it's typically the first command you'd run so we are asking for a stat without really given any value yet to the user.

andresgutgon:
> are we convinced we want to show this on start command? it's typically the first command you'd run so we are asking for a stat without really given any value yet to the user.

Indeed, the tool didn't provide any help at the start but it's also true is a very low-effort request and people might want to help. So I think it's a nice opportunity to let them know they can help with their star.

geclos (commented):
(approved)

PR #376, opened 2024-05-06
Docs / Added width and height attributes to visualizations

samulatitude:
## Describe your changes

Added width and height attributes to visualizations config in docs

### DIFF
--- a/docs/snippets/charts/common.mdx
+++ b/docs/snippets/charts/common.mdx
@@ -62,6 +62,14 @@ To configure and customize the chart, you can add properties, some of which are 
       [See the section yFormat](#axis-x-and-y-options) to learn more about the options.
   </ParamField>
 
+  <ParamField path="height" type="number">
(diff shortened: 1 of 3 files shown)


docs/snippets/charts/common.mdx
```diff
@@ -62,6 +62,14 @@ To configure and customize the chart, you can add properties, some of which are
       [See the section yFormat](#axis-x-and-y-options) to learn more about the options.
   </ParamField>
 
+  <ParamField path="height" type="number">
```
andresgutgon, line 65: Maybe put the attribute in code example to explain that it has to be like this `height={200}`

andresgutgon: If not in all visualizations at least in one https://github.com/latitude-dev/latitude/blob/862c630289310e193ad8e82ba1db71660229d983/docs/views/components/visualizations/area-chart.mdx#syntax

Another option is to put here an example. People not familiar with Svelte syntax can feel confused otherwise. 

```javascript
<AreaChart height={300} />
```

geclos (approved):

Issue #369, opened 2024-05-03
Cloud Embedded Storage layer


andresgutgon:
# What is this?
We want to offer users the possibility of replicating the data from users' databases (sources) into [Parquet files](https://parquet.apache.org/) saved to an S3 bucket. Then all subqueries are done with a [DuckDB](https://duckdb.org/) client.

The benefit of this is that users' DB's receive less load and parquet/duckdb are pretty fast-loading queries. 

### How it works.

Users define a query as a embedded layer by defining some parameters (TO BE DEFINED) in [query's config](https://docs.latitude.so/queries/advanced/config). This configuration tells our embbeding layer 

1. That this query has to be stored in S3 as a Parquet file
2. Every how many days/hours the data is fetched from original database

After this config is in place and query is stored in the Embedded storage layer users can [reference this query](https://docs.latitude.so/queries/advanced/reference-queries) as they would do if the query was done from their database. Under the hood Latitude app will go and fetch this info from S3

## Considerations
Using the embedded layer has some considerations to be made. 

a) All queries referencing this query has to use [DuckDB SQL syntax](https://duckdb.org/docs/sql/query_syntax/select) not users's DB SQL syntax.
b) We need to provide in the queries a metadata saying when was last time this query was updated with original source.
c) How we do the check for periodic updates? I think we need some kind of cron job that pass and check what 
```
bash workspaces -> apps -> queries-in-those-apps
```
need to be stored in Parquet and and need to be refreshed. I think this system has to be a piece a part from current latitude server that run the queries. If we want to go this path we need to access that `queries<-from-app<--from-workspace` somehow from that service. We should start storing the apps reference. 

### TODO
- [x] Decide where lives the infra responsible of fetching the queries and storing it as Parquet files.
- [x] Probably we need a way of running cron jobs. Maybe this initiative can be done together with moving deploys to jobs 
- [x] More things?...

andresgutgon:
For Parquet creation/reading this one looks the one that's best maintained 
https://github.com/LibertyDSNP/parquetjs

[Comparison with others](https://npmtrends.com/@dsnp/parquetjs-vs-node-parquet-vs-parquet-vs-parquetjs-vs-parquetjs-lite-vs-parquets)

andresgutgon:
For DuckDB this one is the official Typescript wrapper around the nodejs client 
https://www.npmjs.com/package/duckdb-async

andresgutgon:
I'm reading about 
[...]

PR #384, opened 2024-05-07
Make source manager available to connectors.

andresgutgon:
# What?
For future changes, we need access to `SourceManager` from the connectors. With this change we be able to do `this.source.manager` inside the connectors

## ⚠️ ⚠️ ⚠️ Refactor ⚠️ ⚠️ ⚠️ 

I removed 🔥 `@latitude-data/base-connector` 🔥  and moved the code under `@latitude-data/source-manager`. I did this to avoid cyclic dependencies between all this code that is part of the same thing in the end.

## Issue ticket number and link
Related with https://github.com/latitude-dev/latitude/issues/369

- [x] Pass `SourceManager` to `Source`
- [x] Pass `Source` to connectors when are initialized.
- [x] Everything should work as before


.changeset/sour-ghosts-fry.md
```diff
@@ -0,0 +1,20 @@
...
+"@latitude-data/snowflake-connector": major
+"@latitude-data/bigquery-connector": major
+"@latitude-data/athena-connector": major
+"@latitude-data/duckdb-connector": major
+"@latitude-data/sqlite-connector": major
+"@latitude-data/mssql-connector": major
+"@latitude-data/mysql-connector": major
+"@latitude-data/trino-connector": major
+"@latitude-data/source-manager": major
+"@latitude-data/cli": major
```
csansoon, line 14: Is this a major change for the cli? This does not change the `SourceManager` API, does it?

andresgutgon: No, but it changed a lot. Also, it brings inside test connector and base connector.

apps/server/src/lib/query_service/find_or_compute.ts
```diff
@@ -1,28 +1,26 @@
...
-}
-
 export default async function findOrCompute({
+  source,
   query,
   queryParams,
   force,
-}: Props): Promise<{
+}: {
+  source: Source
```
andresgutgon, line 12: I pass here source so I can use dependency injection in the test and spy on the underlying connector

packages/connectors/test/src/index.ts
```diff
@@ -1,97 +1,3 @@
...
-      fields: [{ name: 'value', type: DataType.String }],
-      rows: rows.map((value) => [value]),
-      rowCount: rows.length,
-    })
-
-    if (this.onRunQuery) this.onRunQuery(compiledQuery, result)
-    return result
-  }
-}
+export default class TestConnector extends TestConnectorInternal {}
```
andresgutgon, line 3: I do this because I want to use test connector from inside `@latitude-data/source-manager`. Because `@latitude-data/test-connector` has as dependency `@latitude-data/source-manager` I can't install it inside the manager. That gives a Cyclic dependency error 💥 

So the solution is to put the implementantion in the manager. Use it internally and expose it as a different package for the rest of the world


I tried to do
```bash
@latitude-data/source-manager/test-connector` 
```
But I failed miserably and I didn't wanted to expend more time in this

packages/source_manager/src/baseConnector/connectorFactory.ts
```diff
@@ -0,0 +1,101 @@
...
+      throw error
+    }
+    throw new ConnectorNotInstalledError(packageName)
+  }
+}
+
+export class InvalidConnectorType extends Error {}
+
+export function getConnectorPackage(type: ConnectorType) {
+  if (type === ConnectorType.TestInternal) return '../testConnector'
```
andresgutgon, line 81: Internally we use in the `source.yml` as `type=internal_test` and we go to find the connector in a local `.ts` file

packages/source_manager/src/manager/index.ts
```diff
@@ -0,0 +1,116 @@
...
+    sourcePath,
+  }: {
+    sourceFile: string
+    sourcePath: string
+  }) {
+    const schema = readSourceConfig(sourceFile)
+    this.instances[sourcePath] = new Source({
+      path: sourcePath,
+      schema,
+      sourceManager: this,
```
andresgutgon, line 113: This is where I assign `SourceManager` to `Source`

packages/source_manager/src/source/index.ts
```diff
@@ -71,4 +78,18 @@ export class Source {
     const connector = await this.connector()
     return await connector.runCompiled(compiledQuery)
   }
+
+  private async connector(): Promise<BaseConnector> {
+    if (!this._connector) {
+      this._connector = await createConnectorFactory({
+        type: this.type,
+        connectorOptions: {
+          source: this,
```
andresgutgon, line 87: This is where I pass `Source` to connector factory so all connectors has access to `Source` now.

packages/source_manager/src/source/index.ts
```diff
@@ -1,26 +1,40 @@
...
+  }: {
+    path: string
+    schema: SourceSchema
+    sourceManager: SourceManager
+    connector?: BaseConnector
+  }) {
     this._path = path
-    this._schema = config
+    this._schema = schema
+    this.manager = sourceManager
```
andresgutgon, line 36: This way now we can get manager from any connector
``` javascript
// Inside a connector code
this.source.manager // Use what you need
```

apps/server/src/lib/query_service/find_or_compute.ts
```diff
@@ -1,28 +1,26 @@
...
+}): Promise<{
   queryResult: QueryResult
   compiledQuery: CompiledQuery
 }> {
-  const { source, sourceFilePath } = await sourceManager.loadFromQuery(query)
   const compiledQuery = await source.compileQuery({
     queryPath: computeRelativeQueryPath({
       queryPath: query,
-      sourcePath: sourceFilePath,
+      sourcePath: source.path,
```
andresgutgon, line 23: @csansoon why do we need `computeRelativeQueryPath`? I didn't go through this but I would have said that `sourceManager` already knew this

apps/server/src/routes/api/queries/[...query]/+server.ts
```diff
@@ -1,15 +1,18 @@
...
 
 export async function GET({ params: args, url }: Props) {
-  const { query } = args
   try {
     const { params, force, download } = await getQueryParams(url)
+    const query = args.query ?? ''
+    const source = await sourceManager.loadFromQuery(query)
     const { queryResult, compiledQuery } = await findOrCompute({
-      query: query ?? '',
+      source,
```
Comment (andresgutgon) [inline apps/server/src/routes/api/queries/[...query]/+server.ts:14]: Now we inject the `source` in the `findOrCompute`. As we do when testing this function

packages/connectors/athena/src/index.ts
```diff
@@ -47,9 +48,10 @@ export default class AthenaConnector extends BaseConnector {
     },
   }
 
-  constructor(rootPath: string, connectionParams: ConnectionParams) {
-    super(rootPath)
+  constructor(options: ConnectorOptions<ConnectionParams>) {
```
andresgutgon, line 51: Let's try to use always object parameters. Is always easier to add or remove things than positional arguments

apps/server/src/lib/query_service/find_or_compute.test.ts
```diff
@@ -1,70 +1,58 @@
...
-          sourceFilePath: QUERIES_DIR,
-        }
-      },
-    },
-  }
+const connector = new TestConnector({
+  source,
+  connectionParams: { fail: false },
 })
+source.setConnector(connector)
```
csansoon, line 20: Why do you have to set the source's connector manually? Shouldn't it do that automatically?

apps/server/src/lib/query_service/find_or_compute.ts
csansoon: The `sourceManager` already knows, but the `source` didn't. The `source` was only expecting a query path from its source to the query file, and nothing else.

apps/server/src/lib/query_service/find_or_compute.ts
csansoon: The `TestConnector` is designed to be able to spy on it without having to do things like this. Just add a mocked fn as `onRunQuery` of the connector's `connectionParams`, and it will run it for you every time it does.

apps/server/src/routes/api/queries/[...query]/server.test.ts
```diff
@@ -1,48 +1,12 @@
...
-            runCompiledQuery: (compiledQuery: CompiledQuery) =>
-              connector.runCompiled(compiledQuery),
-          },
-          sourceFilePath: filePath,
-        }
-      },
-    },
-    QUERIES_DIR,
-  }
-})
```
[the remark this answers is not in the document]
Comment (csansoon) [reply apps/server/src/routes/api/queries/[...query]/server.test.ts:45]: Just mock the `createConnector` method to return the mocked TestConnector, and everything else should work as expected.

apps/server/src/lib/query_service/find_or_compute.test.ts
andresgutgon: Talked offline

packages/source_manager/src/baseConnector/connectorFactory.test.ts
```diff
@@ -0,0 +1,43 @@
...
+  })
+
+  it('fails when connector is not installed', async () => {
+    await expect(
+      createConnectorFactory({
+        type: ConnectorType.Mssql,
+        connectorOptions: { source, connectionParams: {} },
+      }),
+    ).rejects.toThrowError(
+      "Module @latitude-data/mssql-connector is not a valid Latitude connector. Please make sure you have the correct package installed in your project by running 'npm install @latitude-data/mssql-connector'",
```
csansoon, line 26: It should not say "Module [...] is not a valid Latitude connector" when it is, in fact a valid Latitude connector... right? The problem here is just that it is not installed in the project.

packages/source_manager/src/manager/findSourceConfig.ts
```diff
@@ -44,5 +44,5 @@ export default async function findSourceConfigFromQuery({
     currentDir = path.dirname(currentDir)
   }
 
-  throw new SourceFileNotFoundError('Source file not found')
+  throw new SourceFileNotFoundError(`Source file not found at ${fullPath}`)
```
csansoon, line 47: `Source config file not found for query at ${fullPath}` might be better. Saying "not found at ${fullPath}" sound like you're trying to read that query file as a source path, and that's what is failing.

packages/source_manager/src/baseConnector/connectorFactory.test.ts
andresgutgon: Yep, I'll change

packages/source_manager/src/source/index.ts
```diff
@@ -1,26 +1,40 @@
...
-import type QueryResult from '@latitude-data/query_result'
+} from '@/baseConnector/connectorFactory'
+import { SourceSchema } from '@/types'
+import SourceManager from '@/manager'
 
 export class Source {
   private _schema: SourceSchema
   private _path: string
   private _connector?: BaseConnector
+  manager: SourceManager
```
csansoon, line 21: There should be no way to change the `manager` of a `source`. Change it to `readonly` or set it as private and add a getter.

packages/source_manager/src/source/index.ts
```diff
@@ -35,20 +49,13 @@ export class Source {
     return this._path
   }
 
-  get connectorPackageName(): string {
-    return getConnectorPackage(this.type as ConnectorType)
+  // Testing purposes only
+  setConnector(connector: BaseConnector) {
+    this._connector = connector
   }
```
csansoon, line 55: |>:(

andresgutgon: Disapproval face. Talked offline. I made a small concession here by renaming to `__setConnector`

packages/source_manager/tsconfig.json
```diff
@@ -1,8 +1,11 @@
 {
   "extends": "@latitude-data/typescript/base.json",
   "compilerOptions": {
-    "rootDir": "src",
-    "outDir": "dist"
+    "rootDir": "./src",
+    "outDir": "dist",
+    "paths": {
+      "@/*": ["./src/*"]
```
csansoon, line 7: We're using `$` as src in every other connector. I'd change it here too for consistency

csansoon (approved):

apps/server/src/lib/query_service/find_or_compute.test.ts
geclos: and what was the conclusion? 😅

apps/server/src/lib/query_service/find_or_compute.ts
geclos: I agree and dependency injection makes the api worse as now i can pass a query that references a completely different source than the source i pass in the arguments no?

apps/server/src/lib/query_service/find_or_compute.ts
```diff
@@ -43,7 +41,7 @@ export default async function findOrCompute({
     queryResult = await compute()
   } else {
     queryResult =
-      cache.find(request, compiledQuery.config.ttl) || (await compute())
+      cache.find(request, compiledQuery.config?.ttl) || (await compute())
```
geclos, line 44: config can be nil?

apps/server/src/routes/api/queries/[...query]/+server.ts
Comment (geclos) [reply apps/server/src/routes/api/queries/[...query]/+server.ts:14]: I don't agree with this change tbh

packages/cli/src/lib/sync/syncQueries/ensureConnectorInstalled.ts
```diff
@@ -1,7 +1,7 @@
 import fs from 'fs'
 import colors from 'picocolors'
 import config from '$src/config'
-import SourceManager from '@latitude-data/source-manager'
+import { SourceManager } from '@latitude-data/source-manager'
```
geclos, line 4: with is it no longer the default export?

packages/connectors/athena/src/index.ts
```diff
@@ -36,7 +37,7 @@ interface AthenaQueryClientConfig {
 
 export type ConnectionParams = AthenaQueryClientConfig
 
-export default class AthenaConnector extends BaseConnector {
+export default class AthenaConnector extends BaseConnector<ConnectionParams> {
```
geclos, line 40: don't change the default export i'd say

packages/source_manager/src/source/index.ts
geclos, line 55: can't you do a spy instead? `vi.spyOn(source, '_connector').mockReturnValue(whatever)` adding code only for testing purposes is usually a smell

packages/source_manager/src/types.ts
```diff
@@ -1,9 +1,11 @@
...
+import { ConnectorType } from '@/baseConnector/connectorFactory'
 
-export class NotFoundError {
+export class NotFoundError extends Error {
   message: string
 
   constructor(message: string) {
+    super(message)
     this.message = message
   }
```
geclos, line 10: you can remove everything inside this class since you now extend from Error

packages/source_manager/src/types.ts
```diff
@@ -12,7 +14,7 @@ export class QueryNotFoundError extends NotFoundError {}
 export class SourceFileNotFoundError extends NotFoundError {}
```
geclos, line 14: hmm if you are only using NotFoundError to extend SourceFileNotFoundError you can remove NotFoundError and extend from Error here directly

geclos (approved):
I don't agree with injecting source in findOrCompute, the rest looks 👌🏼

apps/server/src/lib/query_service/find_or_compute.test.ts
andresgutgon: The conclusion

Before we were opening `sourceManager` with this mock
image

I changed the internals of `sourceManager` and this test 3 levels up broke. 

Now we inject `source` in the test and the internals are not touch.

The trade-off is that now we have a public method to set the `source` in the `connector`. This `source.setConnector`. We agree to rename this to `__setConnector` to make it more explicit that it's only for testing purposes.

packages/cli/src/lib/sync/syncQueries/ensureConnectorInstalled.ts
andresgutgon: Correct. Now this package export `SourceManage`, `BaseConnector` and `Source` + types

packages/connectors/athena/src/index.ts
andresgutgon: Why? This way `connectionParams` are typed because is a generic

packages/source_manager/src/source/index.ts
andresgutgon: Yes, maybe this is possible and this is not necessary at all. I'll try in the next PR. This is not a blocking change for this one right?

andresgutgon:
I see a general agreement on not doing the `setConnector`. I'll try to do the spy instead

packages/cli/src/lib/sync/syncQueries/ensureConnectorInstalled.ts
geclos: sure but the default export can keep being the connector no? it's what i'd expect at least

PR #390, opened 2024-05-08
Fix: Old data was leaking into chart after update

csansoon:
## Describe your changes
Updating the data loaded into a chart was still using some rows from the old query result, which made no sense.

Before fix:

https://github.com/latitude-dev/latitude/assets/57395395/9b672715-5209-4a5f-9482-a4bd0fd5bc5c

After fix:

https://github.com/latitude-dev/latitude/assets/57395395/b66ce44e-c60a-4960-bc89-c1a1fb827530

### DIFF
--- a/packages/client/svelte/src/lib/internal/ui/chart/_Echarts.svelte
+++ b/packages/client/svelte/src/lib/internal/ui/chart/_Echarts.svelte
@@ -103,7 +103,7 @@
     onWindowResize()
     return {
       update({ options: newOptions }: ChartableProps) {
-        echartsInstance.setOption({ ...options, ...newOptions })
+        echartsInstance.setOption({ ...options, ...newOptions }, { notMerge: true })
       },
       destroy() {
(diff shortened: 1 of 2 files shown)


packages/client/svelte/src/lib/internal/ui/chart/_Echarts.svelte
```diff
@@ -103,7 +103,7 @@
     onWindowResize()
     return {
       update({ options: newOptions }: ChartableProps) {
-        echartsInstance.setOption({ ...options, ...newOptions })
+        echartsInstance.setOption({ ...options, ...newOptions }, { notMerge: true })
```
andresgutgon, line 106: Is this fixed also in the react package? If not maybe we should move this function to the core

geclos: this needs to be a patch in main branch for react package too

PR #403, opened 2024-05-10
Release

geclos:


examples/sample-react/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "example-react",
   "private": true,
-  "version": "0.0.8",
```
geclos, line 4: this is wrong, main branch should never be at same/higher minor/major version than next branch

packages/cli/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@latitude-data/cli",
   "description": "CLI for Latitude",
-  "version": "1.5.0",
```
geclos, line 4: wrong again, main branch higher than next branch

PR #404, opened 2024-05-10
feature: allow cli to work on non-tty envs

geclos:
## Describe your changes
This allows the cli to run commands where user input is not expected. Of course it won't apply to use cases where user input is expected.

https://github.com/latitude-dev/latitude/assets/1948929/08d9841d-4b05-4ffc-869c-8674a068f330

### DIFF
--- a/.changeset/light-roses-begin.md
+++ b/.changeset/light-roses-begin.md
@@ -1,0 +1,5 @@
+---
+"@latitude-data/cli": patch
+---
+
+Allows cli to run on non-tty environments
--- a/packages/cli/src/lib/telemetry/index.ts
+++ b/packages/cli/src/lib/telemetry/index.ts
@@ -1,32 +1,34 @@
 import RudderAnalytics from '@rudderstack/rudder-sdk-node'
+import config from '$src/config'
+import configStore from '$src/lib/configStore'
 import crypto from 'crypto'
 import os from 'os'
 import { select } from '@inquirer/prompts'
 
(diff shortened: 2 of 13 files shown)


.changeset/light-roses-begin.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latitude-data/cli": patch
+---
+
+Allows cli to run on non-tty environments
```
geclos, line 5: remember that from now-on all main branch commits have to be a patch, regardless of semver semantics. Minor and major versions are reserved from next branches.

packages/cli/src/lib/telemetry/index.ts
```diff
@@ -37,83 +39,48 @@ export class Telemetry {
...
 
-  enable() {
-    this.trackWithoutCheck({ event: 'telemetryEnabled' })
-    configStore.set('telemetry.enabled', true)
-  }
+  private async canTrack() {
+    if (this.enabled === undefined) return await this.askPermission()
 
-  disable() {
-    this.trackWithoutCheck({ event: 'telemetryDisabled' })
```
geclos, line 60: hmm I've removed this events but i think we want them no? To compare non tracked users vs trakced users

andresgutgon (dismissed):
Nice

csansoon (dismissed):

packages/cli/src/lib/telemetry/index.ts
andresgutgon: I think it would be nice

PR #413, opened 2024-05-14
Fix: Properly handle object properties on Query Compiler

csansoon:
## Describe your changes
Compiler did not properly handle object properties. Before this commit, you could not:
- Invoke object properties as methods
- Modify object properties values
- Allow optional chaining operator (`?.`)

### Invoke object properties as methods
Some variable types can contain functions that may be useful in the query compilation run-time. A clear example is Dates, where a Date can be passed to a query as a param, and we may want to run date-related methods.

```
{minDate = param('start_date')}  -- Returns a Date object

SELECT *
FROM table
WHERE year >= {minDate.getYear()}   -- You can now run the .getYear() method from the date object!
```

### Modify object properties as methods
You can now also modify the value from object properties such as arrays and hashes.

```
{ results = runQuery('other_query') }
{ results[0].name = 'foo' }  -- Modifying a property from an object

SELECT *
FROM table
WHERE name IN {#each results as row} {row.name} {/each}
```

### Allow optional chaining operator (`?.`)
The optional chaining operator allows you to read the value of a property located deep within a chain of connected objects without having to expressly validate that each reference in the chain is valid.

```
{ results = runQuery('other_query') }

SELECT *
FROM table
WHERE name = {results[0]?.name ?? 'default'}  -- If there are no results, it will return 'default'
```

### DIFF
--- a/packages/connectors/compiler/src/compiler/compiler.test.ts
+++ b/packages/connectors/compiler/src/compiler/compiler.test.ts
@@ -7,8 +7,8 @@ const compileQuery = (query: string) => {
   return compile({
     query,
     resolveFn: async (value: unknown): Promise<string> => `$[[${value}]]`,
-    supportedMethods: {},
     configFn,
+    supportedMethods: {},
   })
 }
 
@@ -70,8 +70,8 @@ describe('parameterisation of interpolated values', async () => {
     const result = await compile({
       query: sql,
       resolveFn,
-      supportedMethods: {},
       configFn,
+      supportedMethods: {},
     })
 
     const matches = result.match(/\[\d+\]/g) || [] // resolved values in order of appearance in the string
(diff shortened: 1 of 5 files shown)


csansoon:
Since now we can do some things like `{results[0]?.name ?? 'default'}`, I was thinking we could also do something like `{param('start_date')?.getYear()}`. This would mean changing the `param` function where, if the parameter does not exist, it should return `undefined` instead of raising an error. What do you think? Would it be better to crash, or just return undefined?

In the case of `SELECT * FROM user WHERE id = {param('user_id')}`, if no user is given that would not fail, but just return an unexpected response. I don't know what's better for the user in this case.
Thoughts?

geclos:
I agree on supporting object properties on param() calls but I don't understand why it means we need to allow param returning undefined.

packages/connectors/compiler/src/compiler/compiler.test.ts
```diff
@@ -1,3 +1,4 @@
+import { T } from 'vitest/dist/reporters-1evA5lom'
```
geclos, line 1: wat

packages/connectors/compiler/src/compiler/compiler.test.ts
```diff
@@ -129,6 +130,69 @@ describe('variable assignment', async () => {
...
+      {foo.a += 2}
+      {foo.b += 3}
+      {foo.a} {foo.b}
+    `
+    const result = await compileQuery(sql)
+
+    expect(result).toBe('$[[3]] $[[5]]')
+  })
+
+  it('fails when nested value does not exist', async () => {
```
geclos, line 146: do* not

packages/connectors/compiler/src/compiler/compiler.test.ts
```diff
@@ -477,3 +541,187 @@ describe('custom methods', async () => {
...
+  it('correctly evaluates member access', async () => {
+    const sql = `
+      {foo = { bar: 'val' } }
+      {foo.bar}
+    `
+    const result = await compileQuery(sql)
+    expect(result).toBe('$[[val]]')
+  })
+
+  it('allows optional chaining operator', async () => {
```
geclos, line 634: why do we allow it here but not [here](https://github.com/latitude-dev/latitude/pull/413/files#diff-4dd3eb9823025300094929b00fc7b8e7db262d45716b6fc987330453194d337eR157)

geclos (approved):

andresgutgon (approved):

packages/connectors/compiler/src/compiler/compiler.test.ts
csansoon: whoops

PR #416, opened 2024-05-14
BREAKING CHANGE: Relative queries 

andresgutgon:
## Describe your changes
Until now we were requiring users to specify the queries in the [ref](https://docs.latitude.so/queries/advanced/reference-queries#code-example) function from the root of the source folder. Example:

```
MY_QUERIES
  - my-source
    - source.yml
    - query.sql
    - another_query.sql
```
Given that folder structure in `query.sql`

### Before 🚫 
```sql
-- MY_QUERIES/my-source/query.sql
SELECT * FROM {ref('my-source/another_query')}
```

### Now ✅ 
```sql
-- MY_QUERIES/my-source/query.sql
SELECT * FROM {ref('./another_query')}
```
Or ✅ 
```sql
-- MY_QUERIES/my-source/query.sql
SELECT * FROM {ref('another_query')}
```
Or ✅ 
```sql
-- MY_QUERIES/my-source/query.sql
SELECT * FROM {ref('/my-source/another_query')}
```

## Issue ticket number and link
https://github.com/latitude-dev/latitude/issues/369

## Dependency
https://github.com/latitude-dev/latitude/pull/414/files

## TODO
- [x] Pass SQL content from query file from source to connector
- [x] Include in the context request that SQL text
- [x] Fix `runQuery` 🐛. I introduced the [bug here](https://github.com/latitude-dev/latitude/pull/384)
- [x] Use in `ref` method the SQL 
- [x] Fix tests
- [x] QA is working the changes
- [x] Do Changeset

PR #397, opened 2024-05-10
BREAKING CHANGE: Relative queries

andresgutgon:
## Describe your changes
Until now we were requiring users to specify the queries in the [ref](https://docs.latitude.so/queries/advanced/reference-queries#code-example) function from the root of the source folder. Example:

```
MY_QUERIES
  - my-source
    - source.yml
    - query.sql
    - another_query.sql
```
Given that folder structure in `query.sql`

### Before 🚫 
```sql
-- MY_QUERIES/my-source/query.sql
SELECT * FROM {ref('my-source/another_query')}
```

### Now ✅ 
```sql
-- MY_QUERIES/my-source/query.sql
SELECT * FROM {ref('./another_query')}
```
Or ✅ 
```sql
-- MY_QUERIES/my-source/query.sql
SELECT * FROM {ref('another_query')}
```
Or ✅ 
```sql
-- MY_QUERIES/my-source/query.sql
SELECT * FROM {ref('/my-source/another_query')}
```

## Issue ticket number and link
https://github.com/latitude-dev/latitude/issues/369

## TODO
- [x] Pass SQL content from query file from source to connector
- [x] Include in the context request that SQL text
- [x] Fix `runQuery` 🐛. I introduced the [bug here](https://github.com/latitude-dev/latitude/pull/384)
- [x] Use in `ref` method the SQL 
- [x] Fix tests
- [x] QA is working the changes
- [x] Do Changeset

PR #417, opened 2024-05-14
Materialized connector inital setup

andresgutgon:
## What?
We want to allow users cache queries from their sources on disk to improve performance. The queries will be stored in parquet format and read with DuckDB. For that reason, this new connector inherits from our existing DuckDB connector. This PR starts the work in this connector to make it usable in development and writing queries to disk.

## Issue ticket number and link
https://github.com/latitude-dev/latitude/pull/416

## Dependency
https://github.com/latitude-dev/latitude/pull/384
https://github.com/latitude-dev/latitude/pull/397

- [x] Pass `Source` to connectors (done in [another PR](https://github.com/latitude-dev/latitude/pull/384))
- [x] Implement `materializedRef`function


packages/source_manager/src/baseConnector/index.ts
```diff
@@ -282,15 +278,11 @@ export abstract class BaseConnector<P extends ConnectorAttributes = {}> {
...
-          )
-        }
+        this.checkQueryNotCompiled(fullSubQueryPath, queriesBeingCompiled)
 
-        await this.ensureSameSource(fullSubQueryPath)
-
-        const compiledSubQuery = await this.source.compileQuery(
+        const refSource =
+          await this.source.manager.loadFromQuery(fullSubQueryPath)
+        const compiledSubQuery = await refSource.compileQuery(
```
andresgutgon, line 285: This is important. Now `runQuery` is cross-source. Meaning you can call a query from another source.

packages/connectors/materialized/src/index.ts
```diff
@@ -0,0 +1,104 @@
...
+            queryConfig: {}, // Subquery config does not affect the root query
+            ranQueries,
+            queriesBeingCompiled,
+          },
+        )
+
+        const materialize = compiledSubQuery.config.materialize_query
+        const isFalse = materialize === false
+
+        if (isFalse) {
```
csansoon, line 68: The `isFalse` variable is not really needed... You can just do `if (!materialize)`

packages/connectors/materialized/src/index.ts
```diff
@@ -0,0 +1,104 @@
...
+            `Query '${fullSubQueryPath}' is not a materialized query. \nYou can have configured {@config materialized_query = false} in the query file. Set it to 'true'`,
+          )
+        } else if (!materialize) {
+          throw new Error(
+            `Query '${fullSubQueryPath}' is not a materialized query. \nYou can configure it by setting {@config materialized_query = true} in the query file.`,
+          )
+        } else if (Object.keys(accessedParams).length > 0) {
+          const usedParams = this.concatenateParams(accessedParams)
+          throw new Error(
+            `'${referencedQuery}' query can not have parameters to filter the SQL query. You're using ${usedParams} params in the query`,
```
csansoon, line 79: Do we want to show the user which params have they used? I think that can be implicit

packages/source_manager/src/baseConnector/index.ts
```diff
@@ -336,4 +328,17 @@ export abstract class BaseConnector<P extends ConnectorAttributes = {}> {
...
+    queriesBeingCompiled: string[],
+  ): void {
+    const queryName = queryPath.replace(/.sql$/, '')
+
+    if (!queriesBeingCompiled.includes(queryName)) return
+
+    throw new Error(
+      'Query reference to a parent, resulting in cyclic references.',
+    )
+  }
```
csansoon, line 343: Let's follow the same format as with other checks and call it `ensureQueryNotCompiled` or `ensureNoCycles`

csansoon (approved):

packages/connectors/materialized/src/index.ts
andresgutgon: I know but I want to let users know when they configured with `false`

packages/connectors/materialized/src/index.ts
andresgutgon: I think it doesn't hurt no?

PR #420, opened 2024-05-14
Feature/latitude views theming

csansoon:
## Describe your changes
Added support for custom theming on Latitude Views, by adding a `theme` attribute to the latitude.json file. Also, in dev mode, it has hot-reload features to see the new theme in real time.

https://github.com/latitude-dev/latitude/assets/57395395/158e93e2-15a4-46b6-9ff4-ed7ccf3f9007

- The CLI now syncs the latitude.json file to `static/.latitude/latitude.json`
- The server reads the file and sends the theme values to the client.
- Both the server and client update the theme value in real time when the file is updated.
- Also added support for specifying the color theme in the React package

## Hot-reloading
The hot-reloading works differently on the client and the server.
- **server**: For the server, the theme is initialized in `hooks.server.ts`, where the whole latitude.json file is read. When running in dev mode, this same file initializes a `chokidar` watcher to update the value of the theme each time this file is updated.
- **client**: For the client, the theme is initialized in `+layout.svelte`, where the server sends its current value as a `LayoutServerData` via `+layout.server.ts`. Svelte magic stuff. When running in dev mode, a Vite plugin watches for changes in the latitude.json file, reads the config, and sends the updated theme to the client via websockets, which then updates the value for the client in `hooks.client.ts`

### DIFF
--- a/.changeset/clever-moles-prove.md
+++ b/.changeset/clever-moles-prove.md
@@ -1,0 +1,5 @@
+---
+"@latitude-data/react": minor
+---
+
+The LatitudeProvider now accepts a "mode" prop to define the color mode of the theme (light / dark / system)
--- a/apps/server/package.json
+++ b/apps/server/package.json
@@ -19,12 +19,12 @@
     "query": "vite-node scripts/run_query/index.ts"
   },
   "devDependencies": {
-    "@latitude-data/test-connector": "workspace:^",
     "@latitude-data/client": "workspace:*",
     "@latitude-data/eslint-config": "workspace:*",
     "@latitude-data/query_result": "workspace:*",
     "@latitude-data/sql-compiler": "workspace:^",
     "@latitude-data/sveltekit-autoimport": "^1.7.1",
+    "@latitude-data/test-connector": "workspace:^",
     "@rollup/pluginutils": "^5.1.0",
     "@sveltejs/adapter-auto": "^3.0.0",
     "@sveltejs/adapter-node": "^4.0.1",
@@ -38,7 +38,6 @@
     "@types/uuid": "^9.0.8",
     "autoprefixer": "^10.4.17",
     "blessed": "^0.1.81",
-    "chokidar": "^3.5.3",
     "jsdom": "^24.0.0",
     "mock-fs": "^5.2.0",
     "postcss": "^8.4.35",
@@ -52,6 +51,7 @@
     "uuid": "^9.0.1",
     "vite": "^5.0.3",
     "vite-node": "^1.3.1",
+    "chokidar": "^3.5.3",
     "vitest": "^1.2.2"
   },
   "dependencies": {
--- a/apps/server/plugins/configWatcher.ts
+++ b/apps/server/plugins/configWatcher.ts
@@ -1,0 +1,43 @@
+import path from 'path'
+import { type HmrContext } from 'vite'
+import fs from 'fs'
+
+declare module 'vite/types/customEvent' {
+  interface CustomEventMap {
+    'config-update': { [key: string]: unknown }
+  }
+}
+
+export default function watchQueriesPlugin() {
+  const configPath = path.resolve(
+    __dirname,
+    '../static/.latitude/latitude.json',
+  )
+  return {
+    name: 'latitude-json-watcher',
+    enforce: 'post',
+    handleHotUpdate({ file, server }: HmrContext) {
+      if (server.config.mode !== 'development') return
+
+      if (file === configPath) {
+        try {
+          const newConfig = JSON.parse(fs.readFileSync(configPath, 'utf8'))
+          // This update will only be sent to the client, not the server. This means
+          // we need to filter out any config values that are not needed on the client
+          // side for security reasons (even though this is only done in dev mode).
+          const clientConfig = {
+            theme: newConfig.theme,
+            themeMode: newConfig.themeMode,
+          }
+          server.ws.send({
+            type: 'custom',
+            event: 'client-config-update', // Handled in hooks.client.ts
+            data: clientConfig,
+          })
+        } catch (error) {
+          throw new Error(`Failed to read and parse latitude.json\n${error}`)
+        }
+      }
+    },
+  }
+}
--- a/apps/server/src/hooks.server.ts
+++ b/apps/server/src/hooks.server.ts
@@ -1,5 +1,33 @@
 import configureShutdown from '$lib/server/shutdown'
 import { Handle } from '@sveltejs/kit'
+import { APP_CONFIG_PATH } from '$lib/constants'
+import fs from 'fs'
+import { get } from 'svelte/store'
+import { config, LatitudeServerConfig } from '$lib/stores/config'
+
+const isDevMode = import.meta.env.MODE === 'development'
+
+function updateConfig({ fallback }: { fallback: LatitudeServerConfig }) {
+  try {
+    const newConfig = JSON.parse(fs.readFileSync(APP_CONFIG_PATH, 'utf-8'))
+    config.set(newConfig)
+  } catch (error) {
+    console.error('Failed to read and parse latitude.json:', error)
+    config.set(fallback)
+  }
+}
+
+async function watchConfig() {
+  const chokidar = await import('chokidar')
+  const configWatcher = chokidar.watch(APP_CONFIG_PATH)
+  configWatcher.on('change', () => {
+    const currentConfig = get(config)
+    updateConfig({ fallback: currentConfig as LatitudeServerConfig })
+  })
+}
+
+updateConfig({ fallback: {} as LatitudeServerConfig })
+if (isDevMode) watchConfig()
 
 configureShutdown()
 
--- a/apps/server/src/lib/stores/config.ts
+++ b/apps/server/src/lib/stores/config.ts
@@ -1,0 +1,32 @@
+import { writable } from 'svelte/store'
+
+export type LatitudeClientConfig = {
+  theme?: { [key: string]: unknown }
+  themeMode?: 'light' | 'dark' | 'system'
+}
+
+export type LatitudeServerConfig = {
+  name: string
+  version: string
+} & LatitudeClientConfig
+
+export type LatitudeConfig = LatitudeClientConfig | LatitudeServerConfig
+
+/**
+ * Filter out unnecessary config values for the client
+ */
+export function fliterClientConfig(
+  config: LatitudeConfig,
+): LatitudeClientConfig {
+  return {
+    theme: config.theme,
+    themeMode: config.themeMode,
+  }
+}
+
+// This config store is initialized and updated differently in the server and client
+// - In the server, it is initialized and updated in hooks.server.ts
+// - In the client, it is initialized in +layout.svelte (with the initial value from
+//   the server passed from +layout.server.ts) and updated in hooks.client.ts via
+//   the configWatcher plugin
+export const config = writable<LatitudeConfig>({})
--- a/apps/server/src/routes/+layout.server.ts
+++ b/apps/server/src/routes/+layout.server.ts
@@ -1,6 +1,11 @@
-import loadToken, { type TokenResponse } from '$lib/loadToken'
-import { type Load } from '@sveltejs/kit'
+import loadToken from '$lib/loadToken'
+import type { LayoutServerLoad } from './$types'
+import { get } from 'svelte/store'
+import { config, fliterClientConfig } from '$lib/stores/config'
 
-export const load: Load<object, TokenResponse> = async ({ url }) => {
-  return loadToken({ url })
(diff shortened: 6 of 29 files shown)


.changeset/tidy-knives-dance.md
```diff
@@ -0,0 +1,7 @@
+---
+"@latitude-data/svelte": minor
+"@latitude-data/client": patch
+"@latitude-data/server": patch
+---
+
+Fixed automatic Dark Mode detection on Svelte package and Latitude Apps
```
andresgutgon, line 7: This is old no?

packages/cli/src/lib/v1.schema.json
```diff
@@ -5,12 +5,15 @@
...
-      "description": "Latitude app version used."
+  "version": {
+    "type": "string",
+    "minLength": 5,
+    "maxLength": 14,
+    "pattern": "^(?:0|[1-9]\\d*)\\.(?:0|[1-9]\\d*)\\.(?:0|[1-9]\\d*)(?:-next\\.(?:0|[1-9]\\d*))?$",
+    "description": "Latitude app version used."
+  },
+    "theme": {
+      "type": "object"
```
andresgutgon, line 16: Ok, is a start xD

.changeset/tidy-knives-dance.md
csansoon: Whoops, yep. This was a TRENECITO to #412 , not just to `canary`

apps/server/src/routes/+layout.server.ts
```diff
@@ -1,6 +1,16 @@
...
-import { type Load } from '@sveltejs/kit'
+import loadToken from '$lib/loadToken'
+import type { LayoutServerLoad } from './$types'
+import { theme } from '$lib/stores/theme'
+import { theme as client } from '@latitude-data/client'
 
-export const load: Load<object, TokenResponse> = async ({ url }) => {
-  return loadToken({ url })
+let currentTheme: client.skins.PartialTheme | undefined
+theme.subscribe((newTheme) => {
```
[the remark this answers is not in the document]
andresgutgon: Yep, subscribing in the server feels weird

apps/server/src/lib/stores/theme.ts
```diff
@@ -0,0 +1,11 @@
+import { writable } from 'svelte/store'
+import { theme as client } from '@latitude-data/client'
+
+export const APP_CONFIG_PATH = 'static/.latitude/latitude.json'
```
andresgutgon, line 4: Can you move this to `src/lib/constants.ts` I'm going to need it too

csansoon: sure, there you go

.changeset/clever-moles-prove.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latitude-data/react": minor
+---
+
+The LatitudeProvider now accepts a "mode" prop to define the color mode of the theme (light / dark / system)
```
geclos, line 5: mode or thememode? i prefer mode honestly

apps/server/package.json
```diff
@@ -62,6 +61,7 @@
     "@latitude-data/source-manager": "workspace:^",
     "@latitude-data/svelte": "workspace:*",
     "cheerio": "1.0.0-rc.12",
+    "chokidar": "^3.5.3",
```
geclos, line 64: shouldn't this be a dev dependency?

apps/server/plugins/configWatcher.ts
```diff
@@ -0,0 +1,39 @@
...
+        }
+        server.ws.send({
+          type: 'custom',
+          event: 'client-config-update', // Handled in hooks.client.ts
+          data: clientConfig,
+        })
+      }
+    },
+  }
+}
```
geclos, line 39: beautiful attention to detail

apps/server/src/hooks.server.ts
```diff
@@ -1,6 +1,30 @@
...
+  } catch (error) {
+    console.error('Failed to read and parse latitude.json:', error)
+    config.set(defaultConfig)
+  }
+}
+
+updateConfig({} as LatitudeServerConfig)
+
+if (import.meta.env.MODE === 'development') {
+  const configWatcher = chokidar.watch(APP_CONFIG_PATH)
```
geclos, line 22: does this break if the file does not exist?

apps/server/src/lib/stores/config.ts
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Filter out unnecessary config values for the client
+ */
+export function fliterClientConfig(
+  config: LatitudeConfig,
+): LatitudeClientConfig {
+  return {
+    theme: config.theme,
+    themeMode: config.themeMode,
```
geclos, line 23: wouldn't have it been better to have a mode prop inside theme?

packages/client/react/src/lib/theme/ThemeProvider.tsx
```diff
@@ -9,35 +10,55 @@ type PartialTheme = client.skins.PartialTheme
...
+  mode = 'light',
 }: ThemeProviderProps) {
+  const [theme, setFullTheme] = useState<Theme>(createTheme(partialTheme))
+  const setTheme = (theme: PartialTheme) => {
+    setFullTheme(createTheme(theme))
+  }
+
+  const prefersColorScheme = usePrefersColorScheme()
+  const systemMode = useMemo(() => {
+    if (['light', 'dark'].includes(prefersColorScheme))
```
geclos, line 29: missing brackets on multiline if

apps/server/plugins/configWatcher.ts
```diff
@@ -0,0 +1,39 @@
...
+    '../static/.latitude/latitude.json',
+  )
+  return {
+    name: 'latitude-json-watcher',
+    enforce: 'post',
+    handleHotUpdate({ file, server }: HmrContext) {
+      if (server.config.mode !== 'development') return
+
+      if (file === configPath) {
+        const newConfig = JSON.parse(fs.readFileSync(configPath, 'utf8'))
```
geclos, line 23: JSON parse can throw, we should gracefully handle it

.changeset/clever-moles-prove.md
andresgutgon: +1 mode

apps/server/src/lib/stores/config.ts
andresgutgon: I like it this way. In the react version reads beautifu
```javascript 
<LatutudeProvider theme={...} mode='dark | system' />
```

packages/client/react/src/lib/theme/ThemeProvider.tsx
andresgutgon: We should put this in Eslint

apps/server/src/hooks.server.ts
andresgutgon: I think the CLI breaks first before running the server no?

apps/server/src/lib/stores/config.ts
```diff
@@ -0,0 +1,32 @@
+import { writable } from 'svelte/store'
+
+export type LatitudeClientConfig = {
+  theme?: { [key: string]: unknown }
+  themeMode?: 'light' | 'dark' | 'system'
```
andresgutgon, line 5: Do we set a default for this? I guess we do

packages/client/react/src/lib/theme/ThemeProvider.tsx
```diff
@@ -9,35 +10,55 @@ type PartialTheme = client.skins.PartialTheme
...
+      return prefersColorScheme as 'light' | 'dark'
```
andresgutgon, line 30: Missing system here no?

packages/client/react/src/lib/theme/ThemeProvider.tsx
```diff
@@ -9,35 +10,55 @@ type PartialTheme = client.skins.PartialTheme
...
+  theme: partialTheme = defaultTheme,
+  mode = 'light',
 }: ThemeProviderProps) {
+  const [theme, setFullTheme] = useState<Theme>(createTheme(partialTheme))
+  const setTheme = (theme: PartialTheme) => {
+    setFullTheme(createTheme(theme))
+  }
+
+  const prefersColorScheme = usePrefersColorScheme()
+  const systemMode = useMemo(() => {
```
andresgutgon, line 28: I think you can type like this and avoid the `as ` typing
``` javascript
const systemMode = useMemo<ThemeMode>(() => {/*...*/}, [...])
```

packages/client/react/src/lib/theme/ThemeProvider.tsx
andresgutgon, line 29: Nick pick. Do a enum or a constant for the theme mode array

packages/client/svelte/src/lib/ui/theme-provider/index.svelte
```diff
@@ -17,6 +18,10 @@
   let partialTheme: $$Props['theme'] = client.skins.defaultTheme
   export { partialTheme as theme }
 
+  let initialMode: $$Props['mode'] = 'light'
+  export { initialMode as mode }
+  $: setMode(initialMode || 'light')
```
andresgutgon, line 23: Here is the default. This shouldn't be necessary tho. It has already the default on line 21

andresgutgon (approved):
Amazing work Carlos! 👏

.changeset/clever-moles-prove.md
csansoon: I thought a `mode` attribute in the latitude.json could be confusing, since these are settings for the whole project. Someone could think this has something to do with the API, or how the queries are fetched or handled...

apps/server/package.json
csansoon: I get this weird error when leaving `chokidar` as a dev dependency:
image

csansoon: Adding it as `external` to the build rollup options does not work either

csansoon: Okay, fixed it by importing it dynamically instead

apps/server/plugins/configWatcher.ts
csansoon: I want it to throw here, this way the user gets to see the following error **in the client**:

image

Not crashing would just do nothing, and the user may not even notice it.

csansoon: I can throw an error with a more specific with the error message though.

csansoon: image

packages/client/react/src/lib/theme/ThemeProvider.tsx
csansoon: No, for some reason the linter thinks that
```ts
if (['light', 'dark'].includes(prefersColorScheme)) {
  return prefersColorScheme
}
```
could return a value other than 'light' and 'dark' ('no-preference' in this case)

apps/server/plugins/configWatcher.ts
geclos: exactly ☝🏼 the user needs to understand it's an error in latitude.json

.changeset/clever-moles-prove.md
geclos: to be clear i was thinking in a mode prop inside the theme prop of latitude json

packages/client/react/src/lib/theme/ThemeProvider.tsx
csansoon: Yeah, I probably did as a single liner, but prettier added a line break since it was long enough.

apps/server/src/hooks.server.ts
csansoon: CLI breaks before the server even starts. That is, of course, if you update your CLI along with the app. If you only update the app but not the CLI, then this will break.

Actually, this is a good place to add a "please update your CLI" error, since that should be the only case where it can fail.

packages/client/svelte/src/lib/ui/theme-provider/index.svelte
csansoon: Yeah... But turns out it is.
Setting the prop as optional lets the linter think the value can be `undefined` too, in addition to the other values.

Not adding the default on the `let initialMode: $$Props['mode'] || 'light'` line throws this linter error:
```
Property `mode` is optional in type `$$Props` but required in type:
{
  theme?: client.skins.PartialTheme
  mode: 'light' | 'dark' | 'system' | 'undefined'
}
```

While, not adding it to the `$: setMode(initialMode || 'light')` line throws this other one:
```
Type `undefined` is not assignable to parameter of type `"light" | "dark" | "system"`.
```

apps/server/src/hooks.server.ts
csansoon: Apparently it does not break. It just doesn't trigger any change.

PR #424, opened 2024-05-14
Easier theme creation

csansoon:
## Describe your changes

Defining a simple theme sometimes can get annoying.

If the user just wants to change the background color from white to gray, they have not only to change `background`, but to change `card` too (every chart uses a card, and since it's the same color as the background by default it is not obvious that it needs to change), and `echarts.backgroundColor` (otherwise the chart background would still be white). Like this:

```json latitude.json
"theme": {
  "background": "#CCC",
  "card": "#CCC",
  "echarts": {
    "backgroundColor": "#CCC"
  }
}
```
👆 This was just to change the background color... It gets worse with `foreground`.

To fix this, now some attributes inherit the value from other attributes. This means that, if the user only defines the `background` color, `card` and `echarts.backgroundColor` will get this value too. These other values can still be defined by the user independently and they will not be overriden.

```json latitude.json
"theme": {
  "background": "#CCC"
}
```

### DIFF
--- a/packages/client/core/src/theme/skins/addDefaultAttributes.ts
+++ b/packages/client/core/src/theme/skins/addDefaultAttributes.ts
@@ -1,0 +1,18 @@
+import { PartialThemeAttributes } from './types'
+import { removeUndefined } from './utils'
+
+export default function addDefaultAttributes(
+  attrs: PartialThemeAttributes,
+): PartialThemeAttributes {
+  const filledAttrs = {
+    card: attrs['background'],
+    'card-foreground': attrs['foreground'],
+    popover: attrs['background'],
+    'popover-foreground': attrs['foreground'],
+  }
+
+  return {
+    ...(removeUndefined(filledAttrs) as PartialThemeAttributes),
+    ...attrs,
+  }
+}
--- a/packages/client/core/src/theme/skins/buildEchartsTheme.ts
+++ b/packages/client/core/src/theme/skins/buildEchartsTheme.ts
@@ -1,0 +1,45 @@
+import { EchartTheme, TailwindAttributes } from './types'
+import { removeUndefined } from './utils'
+
+export default function buildEchartsTheme(
+  attrs: Partial<TailwindAttributes>,
+): Partial<EchartTheme> {
+  // Echarts theme attributes are by default sourced from the regular Latitude theme attributes
+  const echartsTheme = {
+    backgroundColor: attrs['background'],
+    titleColor: attrs['foreground'],
+    subtitleColor: attrs['muted-foreground'],
+    textColor: attrs['foreground'],
+    markTextColor: attrs['primary-foreground'],
+    borderColor: attrs['border'],
+    legendTextColor: attrs['muted-foreground'],
+    kColor: attrs['primary'],
+    kBorderColor: attrs['primary'],
+    kBorderColor0: attrs['destructive'],
+    graphLineColor: attrs['border'],
+    mapLabelColor: attrs['foreground'],
+    mapLabelColorE: attrs['primary'],
+    mapBorderColor: attrs['muted'],
+    mapBorderColorE: attrs['primary'],
+    mapAreaColor: attrs['muted'],
+    mapAreaColorE: attrs['primary-foreground'],
+    toolboxColor: attrs['muted-foreground'],
+    toolboxEmphasisColor: attrs['foreground'],
+    tooltipAxisColor: attrs['border'],
+    timelineLineColor: attrs['secondary-foreground'],
+    timelineItemColor: attrs['muted-foreground'],
+    timelineItemColorE: attrs['muted-foreground'],
+    timelineCheckColor: attrs['background'],
+    timelineCheckBorderColor: attrs['primary'],
+    timelineControlColor: attrs['secondary-foreground'],
+    timelineControlBorderColor: attrs['secondary-foreground'],
+    timelineLabelColor: attrs['secondary-foreground'],
+    datazoomBackgroundColor: attrs['background'],
+    datazoomDataColor: attrs['border'],
+    datazoomFillColor: attrs['muted'],
+    datazoomHandleColor: attrs['border'],
+    datazoomLabelColor: attrs['muted-foreground'],
+  }
+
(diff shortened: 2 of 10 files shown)


packages/client/core/src/theme/skins/buildEchartsTheme.ts
```diff
@@ -0,0 +1,42 @@
...
+    timelineCheckBorderColor: attrs["primary"],
+    timelineControlColor: attrs["secondary-foreground"],
+    timelineControlBorderColor: attrs["secondary-foreground"],
+    timelineLabelColor: attrs["secondary-foreground"],
+    datazoomBackgroundColor: attrs["background"],
+    datazoomDataColor: attrs["border"],
+    datazoomFillColor: attrs["muted"],
+    datazoomHandleColor: attrs["border"],
+    datazoomLabelColor: attrs["muted-foreground"]
+  }
```
csansoon, line 39: Here's the list of values from echarts that are going to be added from the user defined attributes (if they are defined)

packages/client/core/src/theme/skins/buildEchartsTheme.ts
```diff
@@ -0,0 +1,46 @@
...
+    timelineCheckBorderColor: attrs['primary'],
+    timelineControlColor: attrs['secondary-foreground'],
+    timelineControlBorderColor: attrs['secondary-foreground'],
+    timelineLabelColor: attrs['secondary-foreground'],
+    datazoomBackgroundColor: attrs['background'],
+    datazoomDataColor: attrs['border'],
+    datazoomFillColor: attrs['muted'],
+    datazoomHandleColor: attrs['border'],
+    datazoomLabelColor: attrs['muted-foreground'],
+  }
```
csansoon, line 41: Here's the list of echart attributes and where I am getting the values from.

packages/client/core/src/theme/skins/addDefaultAttributes.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export default function addDefaultAttributes(
+  attrs: PartialThemeAttributes,
+): PartialThemeAttributes {
+  const filledAttrs = {
+    card: attrs['background'],
+    'card-foreground': attrs['foreground'],
+    popover: attrs['background'],
+    'popover-foreground': attrs['foreground'],
+  }
```
csansoon, line 11: Here's the list of default theme attributes and where I am getting the values from.

packages/client/core/src/theme/skins/buildEchartsTheme.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  return Object.fromEntries(
+    Object.entries(echartsTheme).filter(([_, v]) => v !== undefined),
+  ) as Partial<EchartTheme>
```
geclos, line 45: you are doing this transformation in several places, dry a bit 🙏🏼

packages/client/core/src/theme/skins/index.ts
```diff
@@ -15,34 +17,51 @@ export function defaultsDeep<T extends Record<string, unknown>>(
...
+    (acc, key) => {
+      if (!(key in defaultObject)) return { ...acc, [key]: partialObject[key] }
+      const value =
+        typeof defaultObject[key] === 'object' &&
+        !Array.isArray(defaultObject[key])
+          ? defaultsDeep(
+              partialObject[key] || {},
+              defaultObject[key] as Record<string, unknown>,
+            )
+          : partialObject[key] ?? defaultObject[key]
```
geclos, line 30: this assignment is very complex to read, i'd suggest you split all the logical operators in several assignments so that it's easier to understand what is happening here.

packages/client/core/src/theme/skins/index.ts
```diff
@@ -15,34 +17,51 @@ export function defaultsDeep<T extends Record<string, unknown>>(
   partialObject: Partial<T>,
   defaultObject: T,
 ): T {
```
geclos, line 19: couldn't this be done with lodash's [merge](https://lodash.com/docs/4.17.15#merge)? Recursive merge of objects is a solved problem and I'd be less code from us to worry about

packages/client/core/src/theme/skins/index.ts
```diff
@@ -15,34 +17,51 @@ export function defaultsDeep<T extends Record<string, unknown>>(
...
+ * Additionally, it builds some of the Echart theme based on the regular theme attributes.
  */
-function completeDarkAttributes(partialTheme: PartialTheme): PartialTheme {
-  const { dark: partialDark, ...light } = partialTheme
-  const dark = defaultsDeep(partialDark || {}, light)
-  return { ...light, dark }
+function completeTheme(partialTheme: PartialTheme): PartialTheme {
+  const { dark: partialDark, ...partialLight } = partialTheme
+  const light = addDefaultAttributes(partialLight)
+  const dark = defaultsDeep(addDefaultAttributes(partialDark || {}), light)
```
geclos, line 44: better to define defaults in the assignment in line 42 ☝🏼  `{ dark: partialDark = {}`

packages/client/core/src/theme/skins/index.ts
```diff
@@ -15,34 +17,51 @@ export function defaultsDeep<T extends Record<string, unknown>>(
...
+
+  const lightEcharts = defaultsDeep(
+    light.echarts || {},
+    buildEchartsTheme(light),
+  )
+  const darkEcharts = defaultsDeep(dark.echarts || {}, buildEchartsTheme(dark))
```
geclos, line 50: same, try to do defaults in assignment if possible

packages/client/core/src/theme/skins/index.ts
csansoon: I tried lodash's `merge` and `defaultsDeep` methods, but none of them validated my requirements. `defaultsDeep` was the most similar, but had some differences I can't remember rn that didn't work out either.

csansoon: Ah yeah, lodash's `defaultsDeep` was mergeing arrays too, which was not what we needed.

PR #427, opened 2024-05-15
fix: muted text

geclos:
## Describe your changes

Muted text was too dim since we were defaulting to the background muted color instead of the foreground.

image

- [x] I have included a recorded video capture of the feature manually tested (screenshot instead)


packages/client/core/src/theme/ui/alert/types/index.ts
```diff
@@ -8,7 +8,7 @@ const TYPES: Record<AlertyType, Type> = {
   muted: {
     normal: {
       background: 'muted',
-      foreground: 'mutedForeground',
+      foreground: 'muted',
```
andresgutgon, line 11: But now the background and foreground are both `muted` no?

geclos: muted color in text is not the same muted color in background

andresgutgon: I see

andresgutgon (approved):

csansoon (approved):

PR #429, opened 2024-05-15
fix: catch mkdir sync errors in cli syncs

geclos:
## Describe your changes

We've detected this error several times so it's better we handle it gracefully and inform the user.

fixes LATITUDE-DATA-CLI-10

- [x] I have included a recorded video capture of the feature manually tested (noop)

### DIFF
--- a/packages/cli/src/lib/sync/shared/syncFiles.test.ts
+++ b/packages/cli/src/lib/sync/shared/syncFiles.test.ts
@@ -1,0 +1,143 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+import syncFiles from './syncFiles' // Adjust the import path accordingly
+import * as fs from 'fs'
+import colors from 'picocolors'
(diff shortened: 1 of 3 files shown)


packages/cli/src/lib/sync/shared/syncFiles.test.ts
```diff
@@ -0,0 +1,143 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+import syncFiles from './syncFiles' // Adjust the import path accordingly
+import * as fs from 'fs'
+import colors from 'picocolors'
+import output from '$src/lib/output'
+
+// Mocking necessary modules
+vi.mock('fs')
```
andresgutgon, line 8: You could use `mock-fs` to make test work with "real" file system instead of mocking it. Not saying to change it now but as in general I found mock-fs a very nice experience

andresgutgon (approved):

PR #434, opened 2024-05-16
Feature: compiler can read metadata without resolving values

csansoon:
## Describe your changes
Some procedures require extracting metadata from the query, like the defined configuration or knowing which functions are being used, without actually having to resolve the query.

The current way to do this is just compiling the query, which will return some additional metadata along with other compiled output. The problem with this approach is that compiling a query is not always free, as some queries may use `runQuery`. This function is executed in compile time, since we need the result to be able to resolve other parts of the query.  Since we need to get some of the query metadata instantly in some cases, we cannot depend on running queries just to read its configs.

To fix this issue, I have divided the current compilation process in two independent functions: `compile` and `readMetadata`:
- `compile` resolves all variables, parameters and functions in a query, which are dynamic and dependent on context and may take some time depending on its content.
- `readMetadata` in the other hand, returns information that is static in the query, like the defined configuration or the methods that are being called inside of it.

In order to read something as "metadata", it must be static and not depend on other values, parameters or functions resolved during run-time. Since I want the config to be part of the query metadata, I had to add these restrictions to the `{@config}` tag:
- Config must be defined using only literal values. This means it cannot use variables or functions.
- Config tags must be in the root level of the query. This means it cannot be used inside `{#if}` or `{#each}` blocks.

---
🚂 TRENECITO.
🚉 Previous station: #439 
🚉 Next station: #444


packages/connectors/compiler/src/compiler/compiler.test.ts
```diff
@@ -1,3 +1,4 @@
+import { T } from 'vitest/dist/reporters-1evA5lom'
```
andresgutgon, line 1: What's this?

packages/connectors/compiler/src/compiler/logic/nodes/index.ts
```diff
@@ -0,0 +1,34 @@
...
+  [NodeType.ConditionalExpression]: resolveConditionalExpression as ResolveFn,
+  [NodeType.Identifier]: resolveIdentifier as ResolveFn,
+  [NodeType.Literal]: resolveLiteral as ResolveFn,
+  [NodeType.LogicalExpression]: resolveBinaryExpression as ResolveFn,
+  [NodeType.ObjectExpression]: resolveObjectExpression as ResolveFn,
+  [NodeType.MemberExpression]: resolveMemberExpression as ResolveFn,
+  [NodeType.SequenceExpression]: resolveSequenceExpression as ResolveFn,
+  [NodeType.UnaryExpression]: resolveUnaryExpression as ResolveFn,
+  [NodeType.UpdateExpression]: resolveUpdateExpression as ResolveFn,
+}
```
andresgutgon, line 34: Super nice

packages/connectors/compiler/src/compiler/compiler.test.ts
csansoon: idk that was automatic :o

packages/connectors/compiler/src/compiler/logic/nodes/index.ts
csansoon: thanks!

andresgutgon (approved):

PR #435, opened 2024-05-16
Write to parquet file a postgresql query

andresgutgon:
## Describe your changes
We want to allow users to cache queries more permanently. We're doing persistent into parquet files. In this PR we introduce the functionality for writing parquet files to the Source Manager. We also introduce the concept of batched queries in our connectors. This means now they can pull all the data from the query in a way that doesn't exhaust the running machine's memory. This can happen with huge queries

## TODO
- [x] Make sure the connector fails if it has not implemented `batchQuery` method
- [x] Implement batchQuery method in postgresql connector.
- [x] Infer query schema. This is needed to build the parquet file.
- [x] Write each batch of rows from the query to a parquet file
- [x] Find a lib that works to write parquetjs  💀. Finally, I picked [@dsnp/parquetjs](https://github.com/LibertyDSNP/parquetjs)

### DIFF
--- a/apps/server/scripts.rollup.config.mjs
+++ b/apps/server/scripts.rollup.config.mjs
@@ -1,0 +1,18 @@
+import typescript from '@rollup/plugin-typescript'
+
+/**
+ * @typedef {import('rollup').RollupOptions} RollupOptions
+ * @type {RollupOptions}
+ */
+export default {
+  input: 'scripts/materialize_queries/index.ts',
+  output: {
+    file: 'scripts/dist/materialize_queries.js',
+    format: 'esm',
+    sourcemap: true,
+  },
+  plugins: [
+    typescript({ tsconfig: './tsconfig.scripts.json', sourceMap: true }),
+  ],
+  external: ['fs', 'ora', '@latitude-data/source-manager', 'node:util'],
+}
--- a/apps/server/scripts/materialize_queries/index.ts
+++ b/apps/server/scripts/materialize_queries/index.ts
@@ -1,0 +1,97 @@
+import fs from 'fs'
+import { DiskDriver, StorageDriver } from '@latitude-data/source-manager'
+import sourceManager from '../../src/lib/server/sourceManager'
+import ora from 'ora'
+import { parseArgs, type ParseArgsConfig } from 'node:util'
+
+type CommandArgs = {
+  debug: boolean
+}
+
+const OPTIONS = {
+  debug: {
+    type: 'boolean',
+    short: 'd',
+  },
+}
+function getArgs(): CommandArgs {
+  const args = process.argv.slice(2)
+  const { values } = parseArgs({
+    args,
+    options: OPTIONS as ParseArgsConfig['options'],
+  })
+  const { debug } = values as CommandArgs
+
+  return { debug: debug ?? false }
+}
+
+function ensureMaterializeDirExists(storage: StorageDriver) {
+  if (!(storage instanceof DiskDriver)) return
+
+  const basePath = storage.basePath
+  if (!fs.existsSync(basePath)) {
+    fs.mkdirSync(basePath, { recursive: true })
+  }
+}
+
+const BATCH_SIZE = 4096
+async function materializeQueries(debug: boolean) {
+  let startTime = 0
+  if (debug) {
+    startTime = performance.now()
+  }
+  const spinner = ora().start()
+  const defaultArgs = debug
+    ? {
+        onDebug: ({ memoryUsageInMb }: { memoryUsageInMb: string }) => {
+          spinner.text = `Memory: ${memoryUsageInMb}`
+        },
+      }
+    : {}
+  try {
+    // TODO: This is faked. Do the logic to get only materializable queries
+    const allQueries = ['postgresql/query']
+    const storage = sourceManager.materializeStorage
+
+    ensureMaterializeDirExists(storage)
+
+    for (const [index, query] of allQueries.entries()) {
+      const status = `${index + 1} of ${allQueries.length} ${query}`
+      if (debug) {
+        console.log(status)
+      } else {
+        spinner.text = status
+      }
+      const url = await storage.writeParquet({
+        ...defaultArgs,
+        queryPath: query,
+        params: {},
+        batchSize: BATCH_SIZE,
+      })
+
+      if (debug) {
+        console.table({ query, batchSize: BATCH_SIZE, url })
+        const endTime = performance.now()
+        const min = Math.floor((endTime - startTime) / 60000)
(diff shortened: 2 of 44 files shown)


packages/source_manager/src/materialize/drivers/StorageDriver.ts
```diff
@@ -22,8 +87,18 @@ export abstract class StorageDriver {
...
     const hash = createHash('sha256')
-    return hash.update(`${sql}__${queryPath}`).digest('hex')
+    return hash.update(`${sql}__${sourcePath}`).digest('hex')
+  }
+
+  private buildParquetSchema(fields: Field[]) {
+    return fields.reduce((schema, field) => {
+      schema[field.name] = {
+        type: mapDataTypeToParquet(field.type),
+        optional: false, // TODO: How to determine if a field is optional? Maybe from the query result?
```
andresgutgon, line 99: Parquet is strict with the schema and requires that all rows passed to it have the same columns/fields. 

In our case using SQL sources this is not possible and optional is fine being always `false`

What do you think?

packages/source_manager/src/parquetjs-lite.d.ts
```diff
@@ -0,0 +1,24 @@
+declare module 'parquetjs-lite' {
```
andresgutgon, line 1: This package is the newest I found. But is two years old and it doesn't even have types in `@types/` so I made these ambient types checking the source code

packages/connectors/databricks/src/index.ts
```diff
@@ -83,8 +83,9 @@ export default class DatabricksConnector extends BaseConnector<ConnectionParams>
     const queryOperation = await session.executeStatement(compiledQuery.sql, {
       namedParameters: compiledQuery.resolvedParams.reduce(
         (acc, param, index) => {
-          return { ...acc, [`var_${index}`]: param.value }
+          return { ...acc, [`var_${index}`]: param.value as DBSQLParameter }
```
andresgutgon, line 86: I don't know why this started complained in this PR. I didn't touch `resolvedParams` afaik

apps/server/scripts.rollup.config.mjs
```diff
@@ -0,0 +1,18 @@
...
+  output: {
+    file: 'scripts/dist/materialize_queries.js',
+    format: 'esm',
+    sourcemap: true,
+  },
+  plugins: [
+    typescript({ tsconfig: './tsconfig.scripts.json', sourceMap: true }),
+  ],
+  external: ['fs', 'ora', '@latitude-data/source-manager'],
+}
```
csansoon, line 18: Could we do the same and export the `run` script?

apps/server/scripts/materialize_queries/index.ts
```diff
@@ -0,0 +1,82 @@
...
+type CommandArgs = {
+  debug: boolean
+}
+
+function getArgs(): CommandArgs {
+  const args = process.argv.slice(2)
+  const debug = args[0] === 'debug'
+
+  return { debug }
+}
```
csansoon, line 15: If we're adding more scripts like these, we gotta start improving the system with `--debug`-like flags, so that they don't just depend on position.

apps/server/scripts/materialize_queries/index.ts
```diff
@@ -0,0 +1,82 @@
...
+        params: {},
+        batchSize: BATCH_SIZE,
+      })
+
+      if (debug) {
+        console.table({ query, batchSize: BATCH_SIZE, url })
+        const endTime = performance.now()
+        const min = Math.floor((endTime - startTime) / 60000)
+        const seconds = ((endTime - startTime) % 60000) / 1000
+        console.log(`Time: ${min}:${seconds} minutes`)
```
csansoon, line 66: Format the `seconds` with leading zeros when needed. Otherwise, when this variable is lower than 10 it's going to print `12:1`, which is pretty ugly for a timestamp

csansoon: Maybe the `minutes` too. This one's just for OCD tho

packages/connectors/databricks/src/index.ts
csansoon: hmmm. We're starting to have problems where types are not correctly detected from the `@latitude-data/source-manager` package. We may have to investigate this further if it's becoming an issue.

packages/connectors/materialized/src/index.ts
```diff
@@ -76,11 +76,11 @@ export default class MaterializedConnector extends DuckdbConnector {
         const storage = await this.source.manager.materializeStorage
         const materializeUrl = await storage.getUrl({
           sql: compiledSubQuery.sql,
-          queryPath: refSource.path,
+          sourcePath: refSource.path,
           queryName: `materializedRef('${referencedQuery}')`,
         })
 
-        return `(read_parquet('${materializeUrl}'))`
+        return `read_parquet('${materializeUrl}')`
```
csansoon, line 83: Why?

packages/source_manager/src/materialize/drivers/StorageDriver.ts
```diff
@@ -22,8 +141,25 @@ export abstract class StorageDriver {
...
   abstract resolveUrl({ filename }: ResolveUrlParams): Promise<string>
 
-  private hashName({ sql, queryPath }: GetUrlParams) {
+  private hashName({ sql, sourcePath }: GetUrlParams) {
     const hash = createHash('sha256')
-    return hash.update(`${sql}__${queryPath}`).digest('hex')
+    return hash.update(`${sql}__${sourcePath}`).digest('hex')
+  }
+
+  private buildParquetSchema(fields: Field[]) {
```
csansoon, line 149: 🙌

csansoon (approved):

packages/connectors/materialized/src/index.ts
andresgutgon: Wasn't working otherwise

apps/server/scripts/materialize_queries/index.ts
andresgutgon: This is hard OCD carlos. Visit a doctor please

apps/server/scripts.rollup.config.mjs
andresgutgon: I tried and I failed 
image

apps/server/scripts/materialize_queries/index.ts
csansoon: The OCD is about the `minutes`, but the `seconds` should always use leading zeros, otherwise it's wrong. Just add `.toString().padStart(2, '0')`

PR #439, opened 2024-05-21
Restructured compiler logic

csansoon:
## Describe your changes
Moved all of the logic for resolving 'logic nodes' from the main compiler/index.ts file, to different files for better organization. This change is made because this main file was getting way too long and complicated.

1. Main file for compile method is no longer >1000 LOC
2. Logic for how compiler interprets custom code is now way more organized and easier to understand.

This new structure works like this:
Each Logic Node has its own file with a `resolve` method that defines how to resolve the value returned by this method. Then, there's a `resolveLogicNode` method that runs the correct `resolve` function based on the node type.

---
🚂 TRENECITO. This PR was needed so that the next PR is not 2000 LOC :)
🚉 Next station: #434


packages/connectors/compiler/src/compiler/compile.test.ts
```diff
@@ -537,7 +537,7 @@ describe('custom methods', async () => {
       })
     const error = await getExpectedError(action, CompileError)
     expect(error.code).toBe('function-call-error')
-    expect(error.message).toBe("Error calling function 'fooFn': \nError bar")
```
andresgutgon, line 540: Is this ok?

csansoon: Yes! The error message is a bit more generic now, but it will still showing the reference to the exact part of the query where it's the failing function when running `latitude run` for debugging purposes.

PR #444, opened 2024-05-21
Read metadata from supported methods

csansoon:
## Describe your changes
In the previous PR, I added a way for the compiler to read some metadata without resolving values or running queries. This approach worked for simple queries. However, it had a limitation: some functions can add crucial information to the query, which in turn affects the metadata. But if these functions aren’t called, we can’t determine their impact on the current query. For example, if a function contains a `ref` function, the `methods` called in the query (which is information relevant to the metadata) should also contain the methods called in the referenced metadata.

To fix this, I have reworked how `supportedMethods` are defined. Now, each `supportedMethod` must define two functions: `resolve` and `readMetadata`, where each function defines how its compiled result, and its metadata..

In addition, I have changed how the methods handle their requirements. Previously, each function checked and handled its requirements manually. Now, I’ve introduced an extra property for ﻿`supportedMethods` called ﻿`requirements`. This way, the compiler takes charge of managing the function’s behavior and requirements. We now have three requirements:
- `interpolationPolicy`: Whether or not the results from this function can be interpolated into the query. Possible values are `allow` (default), `disallow` or `require`.
- `interpolationMethod`: If the result is going to be interpolated into the query, how is it injected? Possible values are `raw` or `parameterized` (default).
- `requireStaticArguments`: Whether or not the function must use only literal values, or if it can use other variables or logic expressions as arguments. it can be either `true` or `false` (default)

The requirements for the `ref` method have changed, since now we need its arguments to be static to be able to read the metadata from the referenced query, and we cannot know what's the referenced query when reading metadata if the function argument is not static.

---
🚂 TRENECITO.
🚉 Previous station: #434


packages/source_manager/src/materialize/drivers/disk/DiskDriver.test.ts
csansoon, line 1: @andresgutgon Changed your tests here. Would like your opinion!

packages/connectors/compiler/src/compiler/compile.test.ts
csansoon, line 1: Most LOC are here. Not much to see here, just adjusted every test with the new API and added some new tests for the supported methods requirement handling.

packages/connectors/compiler/src/compiler/types.ts
```diff
@@ -1,7 +1,18 @@
...
+  interpolationMethod?: 'parameterize' | 'raw'
+
+  // If the method requires static arguments or allows using logic expressions. Default is false
+  requireStaticArguments?: boolean
+}
+export type SupportedMethod = {
+  requirements?: Partial<SupportedMethodRequirements>
+  resolve: Function
+  readMetadata: (args?: unknown[]) => Promise<QueryMetadata> // Note: `readMetadata` will only receive the function's actual arguments when `requireStaticArguments` is true
+}
```
csansoon, line 15: This is how a `SupportedMethod` is defined now

packages/connectors/compiler/src/compiler/types.ts
```diff
@@ -1,7 +1,18 @@
...
+export type SupportedMethodRequirements = {
+  // Whether the results of the method can be interpolated to the query or not. Default is 'allow'
+  interpolationPolicy?: 'allow' | 'disallow' | 'require'
+
+  // How the method should be interpolated into the query. Default is 'parameterize'
+  interpolationMethod?: 'parameterize' | 'raw'
+
+  // If the method requires static arguments or allows using logic expressions. Default is false
+  requireStaticArguments?: boolean
+}
```
csansoon, line 10: These requirements are now handled by the `CallExpression` node from the compiler. This was previously been checked directly within the function definition, which was not really DRY.

packages/source_manager/src/materialize/drivers/StorageDriver.ts
```diff
@@ -22,8 +26,10 @@ export abstract class StorageDriver {
    */
   abstract resolveUrl({ filename }: ResolveUrlParams): Promise<string>
 
-  private hashName({ sql, queryPath }: GetUrlParams) {
+  static hashName({ sqlHash, sourcePath }: GetUrlParams) {
     const hash = createHash('sha256')
-    return hash.update(`${sql}__${queryPath}`).digest('hex')
+    hash.update(sqlHash)
```
andresgutgon, line 31: Why do we need to hash this here now that we receive it from outside?

packages/source_manager/src/materialize/drivers/StorageDriver.ts
```diff
@@ -22,8 +26,10 @@ export abstract class StorageDriver {
...
+    hash.update(sourcePath)
```
andresgutgon, line 32: Maybe the hash should come already from outside with the source path included. What do you think?

packages/source_manager/src/materialize/drivers/disk/DiskDriver.test.ts
```diff
@@ -1,67 +1,69 @@
...
         queryName: 'query.sql',
       })
-      expect(url).toBe(`${BASE_PATH}/${POSTGRESQL_QUERY_PARQUET}`)
+      const url2 = await driver.getUrl({
+        sqlHash,
+        sourcePath: 'queries/postgresql',
+        queryName: 'query.sql',
+      })
+
+      expect(url1).toBe(url2)
```
andresgutgon, line 37: What's testing this? Is the same call to `getUrl` no?

csansoon: oh... yep. Guess I wanted it to have different `queryName` but forgot to change it.

packages/source_manager/src/materialize/drivers/StorageDriver.ts
csansoon: Hmmm. The compiler is the responsible to hash the SQL contents, and it does not know anything about sources. I think the sqlHash should only be about the actual SQL contents. Does that make any sense?

andresgutgon (approved):

PR #463, opened 2024-05-30
New run command interface

csansoon:
## Describe your changes

Changed `run` command from `blessed` to `ink`. This not only fixes our problems with blessed, but also lets us develop the command's interface with React!

Fixes:
- Script couldn't be compiled
- User couldn't select and copy contents from table
- Terminal got laggy when displaying large tables (every row was being rendered)

https://github.com/latitude-dev/latitude/assets/57395395/7fb9974c-25d0-4fc6-82a6-0437417b12f7

- [ ] I have added a human-readable description of the changes for the release notes


apps/server/package.json
```diff
@@ -62,6 +61,7 @@
   "dependencies": {
     "@latitude-data/client": "workspace:*",
     "@latitude-data/custom_types": "workspace:^",
+    "@latitude-data/display_table": "workspace:^",
```
andresgutgon, line 64: Remember to build run script now that we can

csansoon: done :D

andresgutgon (approved):

packages/cli/core/src/commands/run/index.ts
```diff
@@ -15,16 +15,19 @@ async function run(
 ) {
   const watch = opts?.watch || false
   const debug = opts?.debug || false
+  const params =
+    typeof opts?.param === 'string' ? [opts.param] : opts?.param ?? []
 
-  await syncQueries({ watch })
+  await syncQueries({ watch: true })
```
andresgutgon, line 21: 👀 Watch was false before

csansoon: Now, even when the `--watch` tag is not present, you can manually refresh the results by pressing 'R'! In order to do that, I need the server to have an updated version of the queries, which I do by always enabling the CLI sync.

PR #485, opened 2024-06-03
Feature/update cli when updating project

csansoon:
## Describe your changes

Sometimes when updating a Latitude project, some CLI commands stop working unless it is updated too. To mitigate this, now the CLI will ask you if you want to update the CLI too when you update a project to the latest version, and there's also a new CLI version available.

https://github.com/latitude-dev/latitude/assets/57395395/bdc90a7d-534e-4b1a-8f60-7b279ace0f63


packages/cli/core/package.json
```diff
@@ -11,6 +11,7 @@
   "scripts": {
     "rollup": "rollup -c",
     "build": "NODE_ENV=production pnpm run rollup",
+    "build:dev": "pnpm run rollup",
```
andresgutgon, line 14: why this? we have dev

packages/cli/core/src/lib/versionManagement/updateLatitudeCli.ts
```diff
@@ -0,0 +1,11 @@
+import { CLI_PACKAGE_NAME } from '$src/commands/constants'
+// import config from '$src/config'
```
andresgutgon, line 2: 🔥

packages/cli/core/package.json
csansoon: I want a command that sets the CLI dev config, but only runs once, not continuously. 
Normally, I run ﻿`pnpm build` from the root to make sure everything is built and up to date, including packages I’m not currently working on that may have been updated in another commit. However, always having to run `﻿pnpm dev` in the CLI afterwards to make it work is annoying.

csansoon: Even if I have cli's dev running in a terminal, performing a global `build` will build the cli with the build mode, and the running dev mode will not update it automatically (it only updates it when the src files have changed, not the dist), so I have to stop the dev mode and run it again

PR #499, opened 2024-06-07
Minor/improve vscode sql highlighting

csansoon:
## Describe your changes

Improved syntax on Latitude queries when using Latitude's vscode extension, to better visualize the query's logic and code intentions. Now, Latitude blocks will show actual JS syntax instead of SQL, and block brackets are properly highlighted too. Also, logic blocks that can interpolate content into the query are shown in Italics.

### BEFORE vs NOW
<div>

</div>

Note: These colors are from VSCode's default Dark Theme+. Other themes may redefine these colors.

### Full scope of syntax:
image

### DIFF
--- a/.changeset/new-lies-listen.md
+++ b/.changeset/new-lies-listen.md
@@ -1,0 +1,5 @@
+---
+"latitude-vscode": patch
+---
+
+Improved syntax highlighting for LatitudeSQL files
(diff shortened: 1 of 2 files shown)


andresgutgon (approved):
Missing Changeset

PR #509, opened 2024-06-12
CLI: Only ignore deprecation warnings

csansoon:
## Describe your changes

Instead of removing all warning events, I added a middleware that filters out only deprecation warnings, and keeps any other.

### DIFF
--- a/packages/cli/core/src/cli.ts
+++ b/packages/cli/core/src/cli.ts
@@ -5,9 +5,14 @@ import process from 'process'
  * Avoid shwing deprecation warnings to users
  * for internal packages
  */
-if (process.env.NODE_ENV === 'production') {
-  process.removeAllListeners('warning')
-}
+const onWarnings = process.rawListeners?.('warning')
+process.removeAllListeners('warning')
+process.on('warning', (warning) => {
+  if (warning.name === 'DeprecationWarning') {
+    return
+  }
+  onWarnings?.map((fn) => fn(warning))
+})
 
 import buildCommand from './commands/build'
 import cancelCommand from './commands/cloud/cancel'
--- a/.changeset/swift-ligers-nail.md
+++ b/.changeset/swift-ligers-nail.md
@@ -1,0 +1,5 @@
+---
+"@latitude-data/cli": patch
+---
+
+Only ignore deprecation warnings. Any other warning will still be logged


packages/cli/core/src/cli.ts
```diff
@@ -5,9 +5,14 @@ import process from 'process'
  * Avoid shwing deprecation warnings to users
  * for internal packages
  */
-if (process.env.NODE_ENV === 'production') {
-  process.removeAllListeners('warning')
-}
+const onWarnings = process.rawListeners('warning')
```
csansoon, line 8: On Node 21.4.0, in this context `process` has only 1 warning listener, and it's the following function:
```js
function onWarning(warning) {
  if (!disableWarningSet) {
    disableWarningSet = new SafeSet();
    const disableWarningValues = getOptionValue('--disable-warning');
    for (let i = 0; i < disableWarningValues.length; i++) {
      disableWarningSet.add(disableWarningValues[i]);
    }
  }
  if ((warning?.code && disableWarningSet.has(warning.code)) ||
      (warning?.name && disableWarningSet.has(warning.name))) return;

  if (!(warning instanceof Error)) return;

  const isDeprecation = warning.name === 'DeprecationWarning';
  if (isDeprecation && process.noDeprecation) return;
  const trace = process.traceProcessWarnings ||
                (isDeprecation && process.traceDeprecation);
  let msg = `(${process.release.name}:${process.pid}) `;
  if (warning.code)
    msg += `[${warning.code}] `;
  if (trace && warning.stack) {
    msg += `${warning.stack}`;
  } else {
    msg +=
      typeof warning.toString === 'function' ?
        `${warning.toString()}` :
        ErrorPrototypeToString(warning);
  }
  if (typeof warning.detail === 'string') {
    msg += `\n${warning.detail}`;
  }
  if (!trace && !traceWarningHelperShown) {
    const flag = isDeprecation ? '--trace-deprecation' : '--trace-warnings';
    const argv0 = require('path').basename(process.argv0 || 'node', '.exe');
    msg += `\n(Use \`${argv0} ${flag} ...\` to show where the warning ` +
           'was created)';
    traceWarningHelperShown = true;
  }
  const warningFile = lazyOption();
  if (warningFile) {
    return writeToFile(msg);
  }
  writeOut(msg);
}
```

Here's from where I am taking the 'DeprecationWarning' string that is used to identify Node deprecation warnings.

packages/cli/core/src/cli.ts
```diff
@@ -5,9 +5,14 @@ import process from 'process'
  * Avoid shwing deprecation warnings to users
  * for internal packages
  */
-if (process.env.NODE_ENV === 'production') {
-  process.removeAllListeners('warning')
-}
```
andresgutgon, line 10: 👇 Check if `rawListeners` is present in node 18 > node 22. 
Make sure this is public and we can use it. Otherwise check for the existence of the function and fallback to empty array

csansoon: This works on both Node 18 and 22

PR #515, opened 2024-06-14
S3 storage driver

csansoon:
## Describe your changes

WIP!!

### DIFF
--- a/packages/connectors/duckdb/src/index.ts
+++ b/packages/connectors/duckdb/src/index.ts
@@ -20,8 +20,15 @@ import {
   TableData,
 } from 'duckdb-async'
 
+type S3Credentials = {
+  accessKeyId: string
+  secretAccessKey: string
+  region: string
+}
+
 export type ConnectionParams = {
   url?: string
+  s3?: S3Credentials
 }
 
 export class MaterializedFileNotFoundError extends Error {}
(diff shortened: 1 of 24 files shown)


packages/connectors/duckdb/src/index.ts
```diff
@@ -134,6 +145,25 @@ export default class DuckdbConnector extends BaseConnector<ConnectionParams> {
...
+
+  private async connectS3(credentials: ConnectionParams['s3']): Promise<void> {
+    if (!credentials) return
+    const { accessKeyId, secretAccessKey, region } = credentials
+    const query = `
+INSTALL httpfs;
+LOAD httpfs;
+CREATE SECRET (
+  TYPE S3,
+  KEY_ID ${JSON.stringify(accessKeyId)},
```
andresgutgon, line 161: Do you need to stringify this? Is already a string no?

csansoon: This correctly formats the value, wrapping it in colons and escaping whatever needs escaping.

packages/storage_driver/src/drivers/s3/index.ts
```diff
@@ -0,0 +1,172 @@
...
+
+    upload
+      .done()
+      .then(() => pass.emit(CUSTOM_CLOSE_EVENT))
+      .catch((error) => pass.emit('error', error))
+
+    return pass
+  }
+
+  async createReadStream(filepath: string): Promise<Readable> {
```
andresgutgon, line 163: This should be a private method no?

csansoon: No, but it's not actually being used atm. I'll remove it

andresgutgon (approved):

PR #519, opened 2024-06-19
Some adjustments for Storage Driver and SourceManager to run on the cloud

andresgutgon:
## Describe your changes
We want to use some methods from these 2 packages in our Latitude cloud and we have to make it work.

# TODO
- [x] Do specs for resolve secrets function

### DIFF
--- a/packages/source_manager/src/manager/index.ts
+++ b/packages/source_manager/src/manager/index.ts
@@ -19,6 +19,7 @@ import { createHash } from 'crypto'
 import { ParquetWriter } from '@dsnp/parquetjs'
 import { WriteStreamMinimal } from '@dsnp/parquetjs/dist/lib/util'
 import { buildParquetSchema } from './parquetUtils'
+import { Writable } from 'stream'
 
 const MATERIALIZED_DIR_IN_STORAGE = 'materialized'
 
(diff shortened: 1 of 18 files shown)


packages/source_manager/src/manager/index.ts
```diff
@@ -201,7 +202,9 @@ export default class SourceManager {
       const startTime = performance.now()
       const compiled = await source.compileQuery({ queryPath, params: {} })
 
-      const stream = await this.materializedStorage.createWriteStream(filename)
+      const stream = (await this.materializedStorage.createWriteStream(
+        filename,
+      )) as Writable
```
andresgutgon, line 207: I don't understand why `materializedStorage` is typed a `any`. We must be doing something wrong.

packages/storage_driver/src/drivers/base.ts
```diff
@@ -1,9 +1,15 @@
...
 
 export abstract class StorageDriver {
   constructor() {}
 
+  public abstract type: StorageType
+
+  get isS3Driver(): boolean {
+    return this.type === StorageType.s3
+  }
+
```
csansoon, line 12: What is this for? Can't you just compare the type?

andresgutgon: I can, but this is nicer

csansoon: I think the an abstract parent class like this should be agnostic to its children subclasses existence.
This is not as big of a deal as if it was on the BaseConnector though, here's only two subclasses (for now). Just OCD.

csansoon (approved):
Bit of OCD. Everything else is nice 👌

packages/storage_driver/src/drivers/base.ts
andresgutgon: Yep pure ocd xD

PR #520, opened 2024-06-19
Temp parquet files

csansoon:
## Describe your changes

To avoid keeping broken parquet files, or displaying errors while re-materializing queries, the materialization process will create the parquet files in temp directories. Only when the materialization finished successfully, then the temp file will be moved to the final path where it has to be stored.

### DIFF
--- a/packages/storage_driver/src/drivers/s3/index.ts
+++ b/packages/storage_driver/src/drivers/s3/index.ts
@@ -7,6 +7,8 @@ import {
   GetObjectCommand,
   PutObjectCommand,
   HeadObjectCommand,
+  CopyObjectCommand,
+  DeleteObjectCommand,
 } from '@aws-sdk/client-s3'
 import { Upload } from '@aws-sdk/lib-storage'
 import { PassThrough } from 'stream'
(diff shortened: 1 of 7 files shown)


packages/storage_driver/src/drivers/s3/index.ts
```diff
@@ -132,6 +134,25 @@ export class S3Driver extends StorageDriver {
...
 
+  async move(from: string, to: string): Promise<void> {
+    await this.client.send(
+      new CopyObjectCommand({
+        Bucket: this.bucket,
+        Key: to,
+        CopySource: this.bucket + '/' + from,
+      }),
+    )
+    await this.client.send(
```
andresgutgon, line 145: Wait for Copy To succeed before Deleting

andresgutgon: Also reuse `delete` method here

csansoon: It's waiting for the copy to finish, there's an await on the copyObject sent command

andresgutgon: I mean check if the result is successful

csansoon: If the result is not successful it raises an error.

andresgutgon: True, if the first raise it will not run the delete. So just the small dry for reusing the `delete` method you did

andresgutgon (approved):

PR #530, opened 2024-06-25
Some fixes to make cloud deploy work

andresgutgon:
## Describe your changes
☝️ That

### DIFF
--- a/packages/cli/core/src/templates/Dockerfile.ts
+++ b/packages/cli/core/src/templates/Dockerfile.ts
@@ -1,7 +1,8 @@
 import config from '$src/config'
 
 function cliVersion() {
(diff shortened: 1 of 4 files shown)


packages/cli/core/src/templates/Dockerfile.ts
```diff
@@ -1,7 +1,8 @@
 import config from '$src/config'
 
 function cliVersion() {
-  if (config.dev) return 'latest'
+  // FIXME: Revert this change
+  if (config.dev) return '1.11.0-canary.3'
```
andresgutgon, line 5: I'll remove before publishing a non-canary version. Is only for development in the mornorepo itself anyway

PR #537, opened 2024-06-25
Fix cli version checking

csansoon:
## Describe your changes

CLI is not detecting the installed version correctly. This makes the CLI always update to the latest version. The problem is that we're reading the version from the Latitude Project's root package.json, instead of the `.latitude/app` package.json


andresgutgon (approved):
Thanks

PR #544, opened 2024-06-27
Support materialize tables with JSONB fields

andresgutgon:
## Describe your changes
While doing some tests with materialize I saw we don't treat JSONB columns as jsonb so I added a new data type for it


packages/connectors/postgresql/src/index.ts
```diff
@@ -191,6 +191,8 @@ export default class PostgresConnector extends BaseConnector<ConnectionParams> {
       case pgtypes.builtins.TIMESTAMP:
       case pgtypes.builtins.TIMESTAMPTZ:
         return DataType.Datetime
+      case pgtypes.builtins.JSONB:
+        return DataType.JSONB
```
geclos, line 195: we should check the rest of connectors

geclos (approved):

